PR #3, opened 2023-04-11
Add a "mapping" section.

kixelated:
Indicates the high-level relationship between media and the generic transport. This section will be true regardless of the specific encoding used.

@suhasHere @fluffy @huitema @vasilvv @afrind

### DIFF
--- a/draft-law-moq-warpmedia.md
+++ b/draft-law-moq-warpmedia.md
@@ -81,46 +81,46 @@ This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when 
 # Mapping
 Warp is split into two components:
 
-* A generic transport {{MoQTransport}}
-* A media layer on top.
+* A generic transport. {{MoQTransport}}
+* A media transport layered on top.
 
-This section specifies how media is layered on top of the generic transport.
+This section outlines how the media is generically mapped to the transport.
+Each track format follows this same strategy but has different wire encoding.
 
 ## Tracks
 A media broadcast is broken into media tracks.
-Each track consists of an encoded bitstream, typically compressed using a codec, appended while new media is generated.
+Each track consists of an encoded bitstream, typically compressed using a codec, that grows while new media is generated.
 Examples include audio tracks, video tracks, caption tracks, and metadata tracks.
-
 Each media track MUST be sent over a dedicated transport track {{MoQTransport}}.
 
 ## Groups
 A media track is further fragmented into media groups that can be delivered and decoded independently.
 This allows consumers to skip media when joining a broadcast or during congestion.
-
-The encoder is responsible for determing when to create a new group, which may be done on demand or at a configurable interval.
-Depending on the codec, there may be a significant overhead involved with producing a new group (ex. video I-frame).
-The application is responsible for the trade-off between overhead and when media can be skipped.
-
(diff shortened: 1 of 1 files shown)


draft-law-moq-warpmedia.md
```diff
@@ -78,8 +78,54 @@ This document describes version 1 of the media format.
 
 This document uses the conventions detailed in Section 1.3 of {{!RFC9000}} when describing the binary encoding.
 
-# Packaging
+# Mapping
+Warp is split into two components:
```
wilaw, line 82:
```suggestion
Warp Media Format is split into two components:
```

draft-law-moq-warpmedia.md
```diff
@@ -78,8 +78,54 @@ This document describes version 1 of the media format.
...
+## Groups
+A media track is further fragmented into media groups that can be delivered and decoded independently.
+This allows consumers to skip media when joining a broadcast or during congestion.
+Each media group MUST be sent over a dedicated transport group {{MoQTransport}}.
+
+The encoder is responsible for determining when to create a new group, which may be done on demand or at a configurable interval.
+Depending on the codec, there may be a significant overhead involved with producing a new group (ex. video I-frame).
+The application is responsible for this trade-off.
+
+A media group MAY reference other groups, although it is NOT RECOMMENDED as delivery is not guaranteed.
```
wilaw, line 105: Doesn't this contravene the requirement that groups MUST be independently decodable?

draft-law-moq-warpmedia.md
```diff
@@ -78,8 +78,54 @@ This document describes version 1 of the media format.
...
+However this is often necessary when decoding audio, as reinitializing the decoder for each group can cause noticeable artifacts.
 
+## Objects
+A media group is further fragmented into media objects, the smallest transport unit.
+Media may be fragmented further based on the encoding, but that is opaque to the transport.
+Each media object MUST be sent over a dedicated transport object {{MoQTransport}}.
+
+To recap, a transport object consists of an ordered bitstream.
+The contents of object are streamed in order and without gaps.
```
wilaw, line 114: This 'recap' language is good for a presentation, but we don't need it in a spec. The spec does not need to provide background info. It should be normative, with the minimum informative text to make it understandable.

draft-law-moq-warpmedia.md
```diff
@@ -78,8 +78,54 @@ This document describes version 1 of the media format.
...
+However, an object MAY be starved or reset (with an error code) for any reason, typically in response to congestion.
+These properties are similar to a QUIC stream {{QUIC}}.
+
+The encoder is responsible for determining what media to encode in each object.
+The simplest configuration is to deliver each group as a single object, consisting of all samples/frames in dependency order.
+
+More advanced configurations involve splitting the group into multiple objects, fragmenting further to avoid unnecessary head-of-line blocking.
+For example, an object per video frame, or an object per X audio samples.
+Each object MUST be prioritized according to its dependencies, so that the sender will attempt delivery in a decodable order.
```
wilaw, line 123: Should this spec be more deterministic about what these configurations are? Should they be profiles of Warp media? This language is very loose. We should define possible configurations and make them deterministic and interoperable.

PR #20, opened 2023-10-02
Modifies document to reference external CMAF packaging draft.

### LINKED ISSUE #10: CMAF Mapping
As proposed in #4, I think we can simplify the mapping:

```
|---------------------|----------------|----------------|-----------------|
| Warp                | Catalog Object | Payload Group  | Payload Object  |
|--------------------:|:---------------|----------------|-----------------|
| CMAF {{CMAF}}       | CMAF Header    | CMAF Segment   | CMAF Fragment   |
|---------------------|----------------|----------------|-----------------|
```

To clarify:

- There are any number of frames in a chunk (moof/mdat pair).
- There are any number of chunks in a fragment (styp and moof/mdat pairs).
- There are any number of fragments in a segment. (concatenated together)
- A fragment* is independently decodable.

In order to minimize latency, there should be a chunk per frame, which is the same recommendation as LL-DASH.

If the object has a length in the header, then there should be a chunk per fragment too, but I desperately want to avoid that. That way there would be no latency impact for using a single OBJECT for the entire group.

wilaw:
Modifies document to reference external CMAF packaging draft.  Fixes #10 .

### DIFF
--- a/draft-law-moq-warpstreamingformat.md
+++ b/draft-law-moq-warpstreamingformat.md
@@ -86,6 +86,7 @@ WARP delivers CMAF-packaged media bitstreams. This specification references {{CM
 Both CMAF Object mappings {{CMAFpackaging}} Section 4 are supported and a content producer may use either. To identify to consumers which object mapping mode is being used for a given Track, a new track field is defined as per table 1.
 
 Table 1
+
 | Field                   |  Name                  | Required |  Location |  JSON type |      Definition            |
 |:========================|:=======================|:=========|:==========|:===========|:===========================|
 | WARP packaging mode     | warp-packaging         |  yes     |   RT      |  String    | See {{packagingmode}}      |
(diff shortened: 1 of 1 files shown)


draft-law-moq-warpstreamingformat.md
```diff
@@ -79,13 +80,40 @@ This document describes version 1 of the streaming format.
...
-Each group MUST be independently decodeable. Assigning a new group ID to each CMAF Fragment (see {{CMAF}} Sect 6.6.1) meets this requirement.
+
+## Packaging mode {#packagingmode}
+The packaging mode value is defined by Table 2.
+
+Table 2
+| warp-packing field value  |  Condition                         |                                              Explanation                                            |
+|:==========================|:===================================|:====================================================================================================|
+| frag-per-group            | {{CMAFpackaging}} 4.1 is active    |  Each CMAF Fragment is placed in a single MOQT Object and there is one MOQT Object per MOQT Group   |
+| chunk-per-object          | {{CMAFpackaging}} 4.2 is active    |  Each CMAF chunk is placed in a MOQT Object and there is one MOQT Group per CMAF Fragment           |
```
suhasHere, line 101: I feel the wording is confusing here.

wilaw: I'm not sure what is confusing about `Each CMAF Fragment is placed in a single MOQT Object and there is one MOQT Object per MOQT Group`? CMAF Fragments, MOQT Objects and MOQT Groups are all defined entities. Can you propose some language you consider less confusing?

wilaw: Received comment from Suhas via Webex "the above PR looks good Will .. i re-read once more and you are right". Will merge.

PR #40, opened 2024-11-12
Remove leftover CMAF reference

vasilvv:


wilaw (approved):
LGTM - thanks for catching that.

PR #47, opened 2025-02-24
Replace JSON Patch Updates with a simpler custom implementation

### LINKED ISSUE #46: JSON patch may be too flexible for catalog updates
We define catalog updates as arbitrary modifications via JSON patch.  The question I have is: does this actually make sense, or should we constrain the space of potential updates?  There are some updates that generally make sense, e.g. adding a new track, or changing the resolution on a video track.  But what if the publisher does something that typically does not make sense, e.g. patches a video track to be an audio track?  I feel like it would be easier for the client implementers if we limited the space of potential changes to a well-defined list.

wilaw:
Fixes #46 

This PR removes the RFC6902 JSON Patch update mechanism and replaces it with a simpler update mechanism which allows a restricted set of operations:

- add a track
- clone a track
- remove a track


fluffy:
I like this much better than JSON path stuff

suhasHere (approved):

PR #49, opened 2025-04-02
Constraining Group numbering

### LINKED ISSUE #48: Constrain WARP group numbering
The current MOQT draft only requires Group numbers to increment with time and there are requests to remove even that requirement.

To facilitate the media streaming use case, WARP should constrain group numbers to start at 0 and to increment by 1 with advancing time. This scheme allows transmission nodes (relays) to know when an object is missing without having to make a request to origin. It allows ordering without any other out-of-band communication. It also reduces the information that must be stored in the catalog and it simplifies the timeline track.

wilaw:
Fixes #48 .

Groups within a WARP track MUST start at 0 and MUST increment by 1.

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -690,13 +690,17 @@ The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
 of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
 MOQT Stream.
 
+## Group numbering
+The Group Number of the first Group in any WARP track MUST be 0. Each subsequent
+Group MUST increase by 1.
+
 # Timeline track
 The timeline track provides data about the previously published groups and their
 relationship to wallclock time, media time and associated timed-metadata.
 Timeline tracks allow players to seek to precise points behind the live head in
-a live broadcast, or for random access in a VOD asset. A timeline track may also
+a live broadcast, or for random access in a VOD asset. A timeline track might also
 be used to insert events at media times which do not correlate with Object
-boundaries. Timeline tracks are optional. Multiple timeline tracks MAY exist
+boundaries. Timeline tracks are optional. Multiple timeline tracks can exist
 inside a catalog.
 
 ## Timeline track payload
@@ -731,10 +735,10 @@ contains an array of all track names to which the timeline track applies.
 
 ## Timeline track updating.
 The publisher MUST publish a complete timeline in the first MOQT Object of each
-MOQT Group. The publisher MAY publish incremental updates in the second and
-subsequent Objects within each GROUP. Incremental updates only contain timeline
-events since the last timeline Object. Group duration SHOULD not exceed 30
-seconds.
+MOQT Group of a timeline track. The publisher MAY publish incremental updates
+in the second and subsequent Objects within each GROUP. Incremental updates
+only contain timeline events since the last timeline Object. Group duration
+SHOULD not exceed 30 seconds inside a timeline track.
 
 # Workflow
 
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -691,14 +691,18 @@ of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
 MOQT Stream.
 
 ## Group numbering
-The Group Number of the first Group in any WARP track MUST be 0. Each subsequent
-Group MUST increase by 1.
+The Group ID of the first Group in an track MUST begin with an integer which does
+not repeat in the future. One algorithm that satisifes this requirement is for the
+first Group ID in a track to be the creation time of the first Object expressed as
+the rounded number of milliseconds since Unix epoch. The purpose of this requirement
+is to ensure that republishing the same track in the future, after a loss of
+connectivity or an encoder restart, will never result in smaller or duplicate
+Group IDs being produced under the same track name.
 
-If a publishing session is interrupted at Group ID N midway through Object ID M,
-and then the session is later recovered, the publisher MAY resume publishing on
-the same Track Name and Track Namespace starting at Group ID N+1 Object 0. The
-timeline track, if used, should be updated to indicate a discontinuity at Group
-ID N and Object M using a metadata field entry of "WARP.DISCONTINUITY".
+Each subsequent Group ID MUST increase by 1.
+
+If a publisher is able to maintain state across a republish, it MUST signal the gap
+in Group IDs using the MOQT Prior Group ID Gap Extension header.
 
 # Timeline track
 The timeline track provides data about the previously published groups and their
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -694,6 +694,12 @@ MOQT Stream.
 The Group Number of the first Group in any WARP track MUST be 0. Each subsequent
 Group MUST increase by 1.
 
+If a publishing session is interrupted at Group ID N midway through Object ID M,
+and then the session is later recovered, the publisher MAY resume publishing on
+the same Track Name and Track Namespace starting at Group ID N+1 Object 0. The
+timeline track, if used, should be updated to indicate a discontinuity at Group
+ID N and Object M using a metadata field entry of "WARP.DISCONTINUITY".
+
 # Timeline track
 The timeline track provides data about the previously published groups and their
 relationship to wallclock time, media time and associated timed-metadata.
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -691,13 +691,7 @@ of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
 MOQT Stream.
 
 ## Group numbering
-The Group ID of the first Group in an track MUST begin with an integer which does
-not repeat in the future. One algorithm that satisifes this requirement is for the
-first Group ID in a track to be the creation time of the first Object expressed as
-the rounded number of milliseconds since Unix epoch. The purpose of this requirement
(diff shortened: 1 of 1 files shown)


afrind:
Can you walk through a scenario like "the publisher of a track crashes and restarts and wants to keep publishing on the same track"?  With constrained group IDs, I assume it would be:

TRACK_STAUS_REQUEST to get the last group number, then start publishing.

wilaw:
> Can you walk through a scenario like "the publisher of a track crashes and restarts and wants to keep publishing on the same track"? With constrained group IDs, I assume it would be:
> 
> TRACK_STAUS_REQUEST to get the last group number, then start publishing.

Our current draft requires a relay that does not have an active live stream to go back to the publisher for status request, which would be a recursive round-trip here. I think a better approach would be for the publisher to start republishing on a new key frame (group) and then signal a gap, both via https://github.com/moq-wg/moq-transport/pull/922  and also via an updated timeline track. For the timeline track we should introduce the notion of a discontinuity.

afrind:
> Our current draft requires a relay that does not have an active live stream to go back to the publisher for status request, which would be a recursive round-trip here. 

Interesting point.  I'm thinking about how resumable HTTP uploads work - once you POST'ed part of a resource, you can issue a query to the endpoint to ask how much is there.  It would be slightly strange for the "original publisher" to ask the upload endpoint for previous state, but if it did so before ANNOUNCE, it could still conceivably work (eg: the endpoint wouldn't route the TRACK_STAUS_REQUEST back).

> I think a better approach would be for the publisher to start republishing on a new key frame (group) and then signal a gap, both via [moq-wg/moq-transport#922](https://github.com/moq-wg/moq-transport/pull/922) and also via an updated timeline track. For the timeline track we should introduce the notion of a discontinuity.

Would this PR change to reflect that discontinuities might exist in group ID space as well?  What group ID would the publisher use?

wilaw:
Commit https://github.com/moq-wg/warp-streaming-format/pull/49/commits/704c09744daabc44317949cdb2167b32c0ddfad7 address numbering across breaks:

> If a publishing session is interrupted at Group ID N midway through Object ID M,
> and then the session is later recovered, the publisher MAY resume publishing on
> the same Track Name and Track Namespace starting at Group ID N+1 Object 0. The
> timeline track, if used, should be updated to indicate a discontinuity at Group
> ID N and Object M using a metadata field entry of "WARP.DISCONTINUITY".

gwendalsimon:
Let me try to summarize:
- MoQT has now adopted that group_IDs are disconnected from temporal publication. Filters are integer-based, and the underlying assumption is that subscribers know the structure of group_ID in the track they are interested in (this awareness is out of MoQT scope).
- WARP re-introduces both a strong temporal relation in the group_ID structure, and a Timeline track that carries all time-related publication details.

I am OK with the general approach. It may look like a belt and suspenders one, but, on the one hand, strictly increasing group_ID helps the Relay to manage caches, while on the other hand, timeline track helps the final Subscriber to consume the content. 

Yet, I am not a fan of "WARP.DISCONTINUITY" metadata, which somehow means that all previous group_ID have been published in a perfectly continuous periodical basis. Readers may interpret that the temporal synch is even bigger than just group_ID sequential increase.

Should we say that the "Extension header prior group" https://github.com/moq-wg/moq-transport/pull/922 is not relevant at all in WARP?

wilaw:
> Yet, I am not a fan of "WARP.DISCONTINUITY" metadata, which somehow means that all previous group_ID have been published in a perfectly continuous periodical basis. Readers may interpret that the temporal synch is even bigger than just group_ID sequential increase.

I would expect previous group IDs to be published in a reasonably continuous basis. We can define what that means but "perfect" seems a stronger threshold than is necessary. Here's what I was thinking the timeline track might look like after a disconnect and republish

```
MEDIA_PTS,GROUP_ID,OBJECT_ID,WALLCLOCK,METADATA
64000,32,0,1746171037288,
66000,33,0,1746171039288,
66667,33,10,1746171039955,"WARP.DISCONTINUITY"
80000,34,0,1746171053288,
```

Would you prefer something like WARP.GAP or you don't like the whole metadata approach all-together? If the latter, what would be an alternative?

> Should we say that the "Extension header prior group" https://github.com/moq-wg/moq-transport/pull/922 is not relevant at all in WARP?

I was going to use the gap-signaling we just codified, however I realized that the publisher is in control of the group IDs and I could not come up with a scenario under which the publisher could not ensure that they are contiguous. Can you?

gwendalsimon:
> ```
> MEDIA_PTS,GROUP_ID,OBJECT_ID,WALLCLOCK,METADATA
> 64000,32,0,1746171037288,
> 66000,33,0,1746171039288,
> 66667,33,10,1746171039955,"WARP.DISCONTINUITY"
> 80000,34,0,1746171053288,
> ```
> 
> Would you prefer something like WARP.GAP or you don't like the whole metadata approach all-together? If the latter, what would be an alternative?

We don't necessarily need the timeline to explicitly state the disconnection, do we? The media_PTS + group_ID + wallclock information is, by essence, sufficient to inform the Subscriber.

Thinking about it, I have two additional concerns here:
- Today, the spec has a very loose definition of the Timeline metadata. I would suggest we keep it that way. Otherwise, we would need to define a list of reserved metadata names with a proper definition.
- It may be difficult for a Publisher to even realize that it disconnected. Forcing the Publisher to add such a metadata in the Timeline can be challenging.
 
> I was going to use the gap-signaling we just codified, however I realized that the publisher is in control of the group IDs and I could not come up with a scenario under which the publisher could not ensure that they are contiguous. Can you?

We can omit any mention of the gap-signaling code. Or we can mention that it is useless in WARP, but I don't know if such a mention is relevant.

wilaw:
> We don't necessarily need the timeline to explicitly state the disconnection, do we? The media_PTS + group_ID + wallclock information is, by essence, sufficient to inform the Subscriber.

Correct, the client could infer the discontinuity by comparing delta-media time with delta wallclock time. However, imagine a timeline with 12,000 rows. To know if there is a break in the timeline it would need to process all 12,000 rows, whereas if we explicitly signal it, then it is cheaper to find. I could be persuaded either way.

> Today, the spec has a very loose definition of the Timeline metadata. I would suggest we keep it that way. Otherwise, we would need to define a list of reserved metadata names with a proper definition.

What I had in ind was a middle ground - mostly open metadata, but with a single reserved prefix in which we could code WARP specific timeline events. I'm thinking of things like

WARP.AD_START
WARP.AD_END
WARP.SCTE
WARP.DISCONTINUITY.

Another way to structure the metadata would be as key/value pairs. We can then create WARP-reserved keys. 

> We can omit any mention of the gap-signaling code. Or we can mention that it is useless in WARP, but I don't know if such a mention is relevant.

I think we can mention that its not required, to save the packager a tiny bit of work.

gwendalsimon:
> Correct, the client could infer the discontinuity by comparing delta-media time with delta wallclock time. However, imagine a timeline with 12,000 rows. To know if there is a break in the timeline it would need to process all 12,000 rows, whereas if we explicitly signal it, then it is cheaper to find. I could be persuaded either way.

Me too.
I don't know which use case requires the client to identify whether there is a break in the timeline.

> > Today, the spec has a very loose definition of the Timeline metadata. I would suggest we keep it that way. Otherwise, we would need to define a list of reserved metadata names with a proper definition.
> 
> What I had in ind was a middle ground - mostly open metadata, but with a single reserved prefix in which we could code WARP specific timeline events. I'm thinking of things like
> 
> WARP.AD_START WARP.AD_END WARP.SCTE WARP.DISCONTINUITY.
> 
> Another way to structure the metadata would be as key/value pairs. We can then create WARP-reserved keys.

It is my worry. We would enter into a new round of community debates about formal definitions and mandatory/optional keys (somewhat close to CMSD v2 btw). It is another layer of information to carry, besides SEI and Extension info. 

But I can see the benefits and I would not fight against it.

fluffy:
Discussion from WG meeting 

We agreed to normally increment by 1 other than gaps from crashes / disconnects etc. 

We decided not to require start at 0, instead start at an integer that will have a sufficiently incremented after a crash to not use a group number that was previous used. 

An algorithm that can be used for this is taking the time since uniz epoch in ms as the starting group numbers ( assuming the usage will never use more than 1000 groups per second )

wilaw:
@fluffy  - see commit https://github.com/moq-wg/warp-streaming-format/pull/49/commits/48b4b03109d5f84a7e90b8c455e234afec060cc4 for adjustment of starting Group ID.

draft-ietf-moq-warp.md
```diff
@@ -690,13 +690,27 @@ The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
...
 MOQT Stream.
 
+## Group numbering
+The Group ID of the first Group in an track MUST begin with an integer which does
+not repeat in the future. One algorithm that satisifes this requirement is for the
+first Group ID in a track to be the creation time of the first Object expressed as
+the rounded number of milliseconds since Unix epoch. The purpose of this requirement
+is to ensure that republishing the same track in the future, after a loss of
+connectivity or an encoder restart, will never result in smaller or duplicate
+Group IDs being produced under the same track name.
```
suhasHere:
```suggestion
The Group ID of the first Group published in a track at application startup MUST be a unique integer that will not repeat in the future. One approach to achieve this is to set the initial Group ID to the creation time of the first Object in the group, represented as the number of milliseconds since the Unix epoch, rounded to the nearest millisecond. This ensures that republishing the same track in the future, such as after a loss of connectivity or an encoder restart, will not result in smaller or duplicate Group IDs for the same track name. However, this method does not prevent duplication if more than 1000 groups are published per second.
```

suhasHere (commented):
Looking good overall. I had few minor suggestions. please let me know. Thanks

PR #55, opened 2025-05-01
Adding islive and track duration fields to catalog and a new VOD example

### LINKED ISSUE #50: Catalog should differentiate between live and VOD tracks
Currently the catalog does not differentiate a live track (new media data is being constantly added) from a VOD one ( data is complete and will never change going forward). Players need to understand the difference in order to structure their requests (SUBSCRIBE versus FETCH), and also construct their UI and timeline. 

We propose a new track attribute called "Type" with the following allowed string values:
-    "live" - new objects are being added to the track over time.  Content can be retrieved using a SUBSCRIBE action.
-    "vod" - the track is fully available and no new Objects will be added. Content can only be retrieved using a FETCH action. 

We also propose a "Duration" track attribute which can only be present when the Type is VOD.
-   "Duration"  - the duration of the track rounded to the nearest millisecond.

wilaw:
Fixes #50

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -322,7 +322,7 @@ Table 3: Allowed packaging values
 Location: TF    Required: Required  JSON Type: Boolean
 
 True if new Objects will be added to the track.
-False if no new Objects will be added to the track. 
+False if no new Objects will be added to the track.
 
 ### Track label {#tracklabel}
 Location: TF    Required: Optional   JSON Type: String
@@ -450,7 +450,7 @@ MUST only be included inside a Clone tracks {{clonetracks}} object.
 Location: T    Required: Optional   JSON Type: Number
 
 The duration of the track expressed in integer milliseconds. This field MUST NOT
-be included if the isLive {{islive}} field value is false. 
+be included if the isLive {{islive}} field value is false.
 
 ## Delta updates {#deltaupdates}
 A catalog update might contain incremental changes. This is a useful property if
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -323,7 +323,10 @@ Table 3: Allowed packaging values
 Location: T    Required: Required  JSON Type: Boolean
 
 True if new Objects will be added to the track.
-False if no new Objects will be added to the track.
+False if no new Objects will be added to the track. This is sent under two
+possible conditions:
+* the publisher of a previously live track has ended the track.
+* the track is Video-On-Demand (VOD) and was never live.
 
 ### Track label {#tracklabel}
 Location: TF    Required: Optional   JSON Type: String
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -88,7 +88,8 @@ original publisher. WARP specifies how content should be packaged and signaled,
 defines how the catalog communicates the content, specifies prioritization
 strategies for real-time and workflows for beginning and terminating broadcasts.
 WARP also details how end-subscribers may perform adaptive bitrate switching.
-WARP is targeted at real-time and interactive levels of live latency.
+WARP is targeted at real-time and interactive levels of live latency, as well as
+VOD content.
 
 This document describes version 1 of the streaming format.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-warp.md
```diff
@@ -316,6 +318,12 @@ Table 3: Allowed packaging values
 |:================|:==========|:=================|
 | LOC             | "loc"     | See RFC XXXX     |
 
+### Is Live {#islive}
+Location: TF    Required: Required  JSON Type: Boolean
```
gwendalsimon: T and R are defined, but F and C are not (unless I missed it)

draft-ietf-moq-warp.md
```diff
@@ -756,6 +784,45 @@ description.
 
 ~~~
 
+### Time-aligned VOD Audio/Video Tracks
+
+This example shows catalog for a media producer offering VOD (video on-demand)
+non-live content. The content is LOC packaged, and includes time-aligned audio
+and video tracks.
```
gwendalsimon, line 855: The introduction includes a sentence "WARP is targeted at real-time and interactive levels of live latency." We may fix it if WARP also targets VoD and non-live streaming.

draft-ietf-moq-warp.md
```diff
@@ -438,6 +446,12 @@ Location: T    Required: Optional   JSON Type: String
 A string defining the parent track name {{trackname}} to be cloned. This field
 MUST only be included inside a Clone tracks {{clonetracks}} object.
 
+### Track duration {#trackduration}
+Location: T    Required: Optional   JSON Type: Number
+
+The duration of the track expressed in integer milliseconds. This field MUST NOT
```
gwendalsimon, line 495: Do we use milliseconds or the newly defined timescale unit?

draft-ietf-moq-warp.md
```suggestion
Location: T    Required: Required  JSON Type: Boolean
```
Good catch.

draft-ietf-moq-warp.md
wilaw: Milliseconds. The timescale unit is optional and intended for players playing back single track avc3 content who want to avoid sending init segments.

draft-ietf-moq-warp.md
wilaw: The scope section https://moq-wg.github.io/warp-streaming-format/draft-ietf-moq-warp.html#name-scope defines latency support across multiple ranges including "VOD latency - content that was previously produced, is no longer live and is available indefinitely.". The abstract needs to be updated to match the scope.

draft-ietf-moq-warp.md
wilaw: Fixed with https://github.com/moq-wg/warp-streaming-format/pull/55/commits/a9b1dcee93c143d586f2e518d397f4dec1811148

draft-ietf-moq-warp.md
```diff
@@ -316,6 +318,12 @@ Table 3: Allowed packaging values
 |:================|:==========|:=================|
 | LOC             | "loc"     | See RFC XXXX     |
 
+### Is Live {#islive}
+Location: T    Required: Required  JSON Type: Boolean
+
+True if new Objects will be added to the track.
+False if no new Objects will be added to the track.
```
suhasHere:
```suggestion
False if no new Objects will be added to the track. This is typically sent when the publisher has ended the track.
```

suhasHere (commented):
nits .. otherwise LGTM

suhasHere (approved):

draft-ietf-moq-warp.md
wilaw: > This is typically sent when the publisher has ended the track.

This is sent under two "typical" conditions:
1. If the track is VOD 
2. If the track was live and has now ended.

draft-ietf-moq-warp.md
wilaw: Commit https://github.com/moq-wg/warp-streaming-format/pull/55/commits/95a57362fcb6a985669f3fa4ac684fa0cafdbd79 fixes the introductory paragraph to match the scope.

draft-ietf-moq-warp.md
wilaw: Commit https://github.com/moq-wg/warp-streaming-format/pull/55/commits/bc790e8322cae73bbd3996ad2f8be35cb22e309b updates the isLive = false definition.

PR #57, opened 2025-05-01
Adding the track type field and updating examples

### LINKED ISSUE #54: Catalog should define the 'type' of each track
> [7.2. ](https://moq-wg.github.io/warp-streaming-format/draft-ietf-moq-warp.html#section-7.2)[Timeline Catalog requirements](https://moq-wg.github.io/warp-streaming-format/draft-ietf-moq-warp.html#name-timeline-catalog-requiremen)
> A timeline track MUST carry a 'type' identifier in the Catalog with a value of "timeline".

However no 'type' field exists for the catalog. 

The proposal is to add a 'type' field. Value is a JSON string with an enumeration of reserved values

"video" - carries visual content
"audio" - carries audio content
"timeline" - carries a timeline track
"caption" - provides a full textual representation of the audio track for deaf or hard-of-hearing viewers
"subtitle" - a transcription of the spoken dialogue, potentially in an alternate language. 
"audiodescription" - an audio description to promote accessibility for visually impaired users
"signlanguage" - a visual track for hearing impaired users. 

These should be specified in an IANA table so we can add more later, for VR, gaming etc. 

We should also allow custom 'types" to be specified , as long as thet do not collide with the reserved types.

wilaw:
Fixes #54

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -322,10 +322,10 @@ Table 3: Allowed packaging values
 Location: T    Required: Optional   JSON Type: String
 
 A string defining the type of content carried by the track. Reserved types
-are described in Table 4. These type values are case-sensitive. 
+are described in Table 4. These type values are case-sensitive.
 
 This type field MAY be used in conjunction with the Mimetype {{mimetype}} to
-fully describe the content of the track. 
+fully describe the content of the track.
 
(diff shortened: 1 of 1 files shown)


suhasHere:
Conclusion from the May 2025 Interim
  - Rename type to either intent/role/description
  - Separate description from the actual media being carried

PR #61, opened 2025-07-12
Adding isComplete and rules for terminating a broadcast

### LINKED ISSUE #59: How do we signal the end of a broadcast?
A SUBSCRIBE_DONE can be sent for each active subscription. However, the tracks may begin and end at different times, or more tracks may be published in the future.

What is the signal to indicate that the client can unsubscribe from the catalog track because the broadcast is complete?

Some options:
1. Send a catalog with no tracks (this does not solve for the case that there may be future tracks made available)
2. Create an explicit catalog field for "END OF BROADCAST".
3. Some other signal?

wilaw:
This PR adds a definition for the isComplete catalog field, adds a new catalog example and adds a workflow section instructing publishers how to cleanly end a broadcast.

Fixes #59

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -298,7 +298,8 @@ Location: R    Required: Optional    JSON Type: Boolean
 
 Issued once a previously live broadcast is complete. This is a commitment that all
 tracks are complete, no new tracks will be added and no new content will be
-published. This field MUST NOT be included if it is FALSE.
+published. This field MUST NOT be included if it is FALSE. This field MUST NOT be
+removed from a catalog once it has been added.
 
 ### Tracks {#tracks}
 Location: R    Required: Yes    JSON Type: Array
@@ -972,8 +973,8 @@ A timeline track MUST carry a 'type' identifier in the Catalog with a value of
 contains an array of all track names to which the timeline track applies.
 
 ## Timeline track updating.
-The publisher MUST publish a complete timeline in the first MOQT Object of each
-MOQT Group of a timeline track. The publisher MAY publish incremental updates
+The publisher MUST publish an indepdendent timeline in the first MOQT Object of
+each MOQT Group of a timeline track. The publisher MAY publish incremental updates
 in the second and subsequent Objects within each GROUP. Incremental updates
 only contain timeline events since the last timeline Object. Group duration
 SHOULD not exceed 30 seconds inside a timeline track.
@@ -986,13 +987,17 @@ track objects.
 
 ## Ending a live broadcast
 After publishing a catalog and defining tracks carrying live content, an original
-publisher can deliver a deterministic signal to all subscribers that it is complete
-by taking the following steps:
+publisher can deliver a deterministic signal to all subscribers that the broadcast
+is complete by taking the following steps:
 
 * Send a SUBSCRIBE_DONE (See MOQT Sect 8.1.2) message for all active tracks using
   status code 0x2	Track Ended.
-* Publish a catalog update, either absolute or as delta update, which signals
-  isComplete as TRUE.
+* If the live stream is being converted instantly to a VOD asset, then publish an
+  independent (non-delta) catalog update which, for each tracks, sets isLive {{islive}}
+  to FALSE and adds a track duration {{trackduration}} field.
+* If the live stream is being terminated permanently without conversion to VOD, then
+  publish an independent catalog update which signals isComplete {{iscomplete}} as
+  TRUE and which contains an empty Track {{tracks}} field.
 
 # Security Considerations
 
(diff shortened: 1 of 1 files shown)


yekuiwang (changes requested):
In 

> Issued once a previously live broadcast is complete. This is a commitment that all
> tracks are complete, no new tracks will be added and no new content will be
> published. This field MUST NOT be included if it is FALSE.
> 

Change "This field MUST NOT be included if **it** is FALSE." to "This field MUST NOT be included if **the isLive field** is FALSE."

draft-ietf-moq-warp.md
```diff
@@ -292,6 +293,13 @@ The wallclock time at which this catalog instance was generated, expressed as th
 number of milliseconds that have elapsed since January 1, 1970 (midnight UTC/GMT).
 This field SHOULD NOT be included if the isLive field is false.
 
+### Is Complete {#iscomplete}
+Location: R    Required: Optional    JSON Type: Boolean
+
+Issued once a previously live broadcast is complete. This is a commitment that all
+tracks are complete, no new tracks will be added and no new content will be
+published. This field MUST NOT be included if it is FALSE.
```
```suggestion
published. This field MUST NOT be included if its value is FALSE.
```

yekuiwang: I see, you don't want include it to indicate that it is not yet complete.

However, when not included, the field does not exist, and then saying the value of the field is a bit weird. Also, you don't want to include it when the value of the isLive field is FALSE.

How about the following: "When this field is included, its value MUST be TRUE and the value of the isLive field MUST be TRUE."

wilaw: Actually, I'm not sure that the value of the isLive field in the tracks matters. IN the example I added for termination, I omitted the isLive field and all track defintions. The isComplete field is at the root level of the catalog, meaning it applies to all tracks in the broacast, not just one track. The isLIVE field is a per track property. Since the broadcast is complete, strictly speaking none of the tracks exists anymore.  It also not VOD, its just not available anymore at all. 

I want to talk through what should happen if I have a live broadcast that I want to terminate and immediately make it available as VOD. This is quite possible with a larger enough cache. Should I:
1. Publish the track as isLive=true, then send isComplete=true with isLive=true, then update catalog to say isLive=false along with duration=XX, or
2. Publish the track as isLive=true, then immediately then update catalog to say isLive=false along with duration=XX.

So I'm beginning to think that there should be 2 ways to end a live track
1. If it is not being converted to VOD, then issue isComplete=true and omit all track definitions
2. if it is being converted to VOD, then simply change to isLive =false and add a duraiton.

My feeling with isComplete is that you are guaranteeing that there will be no more catalog updates, including one that might switch the broadcast to VOD. So my preference would be the workflow of [2].

yekuiwang: Got it. I did not consider the use case of turning a live streaming into VOD. Also I did not pay attention that isLive is a track-level field; good to learn that a broadcast can be a mix of live and VOD tracks.

No more comment herein. Please feel free to go head to merge this PR.

draft-ietf-moq-warp.md
```diff
@@ -889,6 +898,20 @@ and video tracks.
...
+### Terminating a live broadcast
+
+This example shows catalog for a media producer terminating a previously
+live broadcast containing a video and an audio track.
+
+~~~json
+{
+  "version": 1,
+  "generatedAt": 1746104606044,
+  "isComplete": TRUE
```
```suggestion
  "isComplete": TRUE,
  "tracks": []
```

draft-ietf-moq-warp.md
```diff
@@ -962,10 +986,19 @@ SHOULD not exceed 30 seconds inside a timeline track.
...
-
+## Ending a live broadcast
+After publishing a catalog and defining tracks carrying live content, an original
+publisher can deliver a deterministic signal to all subscribers that the broadcast
+is complete by taking the following steps:
+
+* Send a SUBSCRIBE_DONE (See MOQT Sect 8.1.2) message for all active tracks using
+  status code 0x2	Track Ended.
+* If the live stream is being converted instantly to a VOD asset, then publish an
+  independent (non-delta) catalog update which, for each tracks, sets isLive {{islive}}
```
```suggestion
  independent (non-delta) catalog update which, for each track, sets isLive {{islive}}
```

draft-ietf-moq-warp.md
```diff
@@ -962,10 +986,19 @@ SHOULD not exceed 30 seconds inside a timeline track.
...
+is complete by taking the following steps:
+
+* Send a SUBSCRIBE_DONE (See MOQT Sect 8.1.2) message for all active tracks using
+  status code 0x2	Track Ended.
+* If the live stream is being converted instantly to a VOD asset, then publish an
+  independent (non-delta) catalog update which, for each track, sets isLive {{islive}}
+  to FALSE and adds a track duration {{trackduration}} field.
+* If the live stream is being terminated permanently without conversion to VOD, then
+  publish an independent catalog update which signals isComplete {{iscomplete}} as
+  TRUE and which contains an empty Track {{tracks}} field.
```
```suggestion
  TRUE and which contains an empty Tracks {{tracks}} field.
```

PR #62, opened 2025-07-29
Some changes agreed on the MoQ mailing list

### LINKED ISSUE #63: On the requirement of alternate tracks having matching framerate
Section 5.2.15: It is said that alternate tracks SHOULD have matching framerate Section 5.2.22 and media time sequences. 
 
Does “matching framerate” herein mean the same framerate? If so, even the wording uses SHOULD not MUST, it is probably too strong a requirement, as it is so typical in adaptive streaming to have alterative representations of the same video content with different framerates, as for some video content when the spatial resolution is low it simply does not make sense from rate-distortion optimization point of view to have a high framerate value. Therefore, IMHO it is better not to have this requirement.

yekuiwang:
This PR includes all changes recently agreed on the MoQ mailing list to draft-ietf-moq-warp-01, between Will and Ye-Kui.

Fixes: #63

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -105,7 +105,7 @@ The purpose of WARP is to provide an interoperable media streaming format
 operating over {{MoQTransport}}. Interoperability implies that:
 
 * An original publisher can package incoming media content into tracks, prepare
-  a catalog and annouce the availability of the content to an MOQT relay. Media
+  a catalog and announce the availability of the content to an MOQT relay. Media
   content refers to audio and video data, as well as ancillary data such as
   captions, subtitles, accessibility and other timed-text data.
 * An MOQT relay can process the annoucement as well as cache and propagate the
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -949,13 +949,15 @@ Each timeline track begins with a header row of MEDIA_PTS,GROUP_ID,OBJECT_ID,
 WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
 
 * MEDIA_PTS: a media timestamp rounded to the nearest millisecond. This entry
-  MUST NOT be empty. If the Object ID entry is present, then this value MUST
-  match the media presentation timestamp of the first media sample in the
-  referenced Object. Otherwise, this value MUST match the media presentation
-  timestamp of the first media sample in an Object that is in the Group but
-  is not the first Object in the Group. 
+  MUST NOT be empty. If the Object ID entry is present and equal to -1, this
+  value MUST match the media presentation timestamp of the first media sample
+  in an Object that is in the Group but is not the first Object in the Group.
+  Otherwise, this value MUST match the media presentation timestamp of the
+  first media sample in the referenced Object.
 * GROUP_ID: the MOQT Group ID. This entry MUST not be empty.
-* OBJECT_ID: the MOQT Object ID. This entry MAY be empty.
+* OBJECT_ID: the MOQT Object ID. This entry MAY be empty. If this entry is
+  present and equal to -1, the Object ID is unknow. Otherwise if this entry
+  is absent, the the Object ID is 0.
 * WALLCLOCK: the wallclock time at which the media was encoded, expressed as
   the number of milliseconds that have elapsed since January 1, 1970
   (midnight UTC/GMT). For VOD assets, or if the wallclock time is not known,
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -419,7 +419,7 @@ Location: T    Required: Optional   JSON Type: Number
 
 A number identifying the temporal layer/sub-layer encoding of the track,
 starting with 0 for the base layer, and increasing by 1 for the next higher
- temporal fidelity.
+temporal fidelity.
 
 ### Spatial ID {#spatialid}
 Location: T    Required: Optional   JSON Type: Number
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -972,7 +972,7 @@ A timeline track MUST carry a 'type' identifier in the Catalog with a value of
 "timeline". A timeline track MUST carry a 'dependes' attribute which
 contains an array of all track names to which the timeline track applies.
 
-## Timeline track updating.
+## Timeline track updating
 The publisher MUST publish a complete timeline in the first MOQT Object of each
 MOQT Group of a timeline track. The publisher MAY publish incremental updates
 in the second and subsequent Objects within each GROUP. Incremental updates
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -105,10 +105,10 @@ The purpose of WARP is to provide an interoperable media streaming format
 operating over {{MoQTransport}}. Interoperability implies that:
 
 * An original publisher can package incoming media content into tracks, prepare
-  a catalog and annouce the availability of the content to a MOQT relay. Media
+  a catalog and annouce the availability of the content to an MOQT relay. Media
   content refers to audio and video data, as well as ancillary data such as
   captions, subtitles, accessibility and other timed-text data.
-* A MOQT relay can process the annoucement as well as cache and propagate the
+* An MOQT relay can process the annoucement as well as cache and propagate the
   tracks, both to other relays or to the final subscriber.
 * A final subscriber can parse the catalog, request tracks, decode and render
   the received media data.
@@ -133,7 +133,7 @@ that end, the following features are within scope:
  * VOD latency - content that was previously produced, is no longer live and is
    available indefinitely.
 * Content encryption
-* ABR between time-synced tracks - subscribers may switch between between tracks
+* ABR between time-synced tracks - subscribers may switch between tracks
   at different quality levels in order to maximize visual or audio quality under
   fconditions of throughput variability.
 * Capable of delivering interstitial advertising.
@@ -161,8 +161,8 @@ When LOC packaging is used for a track, the catalog packaging attribute
 WARP Tracks MAY be time-aligned. Those that are, are subject to the following
 requirements:
 
-* Time-aligned tracks MUST be advertised in the catalog as belonging to a common
-  render group.
+* Tracks advertised in the catalog as belonging to a common render group MUST
+  be time-aligned.
 * The render duration of the first media object of each equally numbered MOQT
   Group, after decoding, MUST have overlapping presentation time.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-warp.md
```diff
@@ -105,10 +105,10 @@ The purpose of WARP is to provide an interoperable media streaming format
 operating over {{MoQTransport}}. Interoperability implies that:
 
 * An original publisher can package incoming media content into tracks, prepare
-  a catalog and annouce the availability of the content to a MOQT relay. Media
+  a catalog and annouce the availability of the content to an MOQT relay. Media
```
```suggestion
  a catalog and announce the availability of the content to an MOQT relay. Media
```

draft-ietf-moq-warp.md
```diff
@@ -105,10 +105,10 @@ The purpose of WARP is to provide an interoperable media streaming format
...
   content refers to audio and video data, as well as ancillary data such as
   captions, subtitles, accessibility and other timed-text data.
-* A MOQT relay can process the annoucement as well as cache and propagate the
+* An MOQT relay can process the annoucement as well as cache and propagate the
```
```suggestion
* An MOQT relay can process the announcement as well as cache and propagate the
```

draft-ietf-moq-warp.md
```diff
@@ -932,8 +932,10 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
 * MEDIA_PTS: a media timestamp rounded to the nearest millisecond. This entry
   MUST NOT be empty. If the Object ID entry is present, then this value MUST
   match the media presentation timestamp of the first media sample in the
-  referenced Object.
-* GROUP_ID: the MOQT Group ID. This entry MAY be empty.
+  referenced Object. Otherwise, this value MUST match the media presentation
+  timestamp of the first media sample in an Object that is in the Group but
+  is not the first Object in the Group. 
+* GROUP_ID: the MOQT Group ID. This entry MUST not be empty.
```
wilaw, line 951: What about the case of metadata entries , such as this proposed Event:

https://docs.google.com/presentation/d/1MgcASdH_b8gUpMHjO-05Gbk_0CP8yw0RQO9gVO0UHrg/edit?slide=id.g21bc8d0f159_0_23#slide=id.g21bc8d0f159_0_23

It seems that the media time MUST be present, but a Group ID is not required. So I would revert it to "MAY be empty".

wilaw (commented):
LGTM except for the suggested spelling corrections (which were my own :() and one comment on GROUP ID being required.

draft-ietf-moq-warp.md
gwendalsimon: We can also have the case of a metadata attached to an object but the origin of the timeline is not sure about the Media_PTS. For example an external tool analyzes the first Object in a group (an IDR frame) and returns a metadata to add to the timeline. It may only know the frame but nothing else. Can it be possible?
If so, I would say that at least one of the (MEDIA_PTS, GROUP_ID) MUST be present.

yekuiwang:
See this email of mine on the MoQ mailing list: https://mailarchive.ietf.org/arch/msg/moq/VdcK-w_2vXkcxd227QI6_HhevRE/

yekuiwang:
This new commit named "Requirement on alternative tracks" fixes #63.

draft-ietf-moq-warp.md
yekuiwang: @wilaw As I said in the reflector email mentioned earlier, I am OK with @gwendalsimon's suggestion "at least one of the (MEDIA_PTS, GROUP_ID) MUST be present." Will are you OK with that? If so, then this aspect can be concluded and I can provide a commit for the text change. 

Also, please confirm or express your opinion otherwise regarding whether to have Object ID. Herein, it seems that based on the email reflector discussion, we are going to keep the presence but update the semantics such that -1 means unknown, and absence means 0. Again, once confirmed, I can provide the text change in a commit.

wilaw: @yekuiwang .

I agree with the statement that "at least one of the (MEDIA_PTS, GROUP_ID) MUST be present.". As a consequence of accepting this, I have a concern about ordering. Currently the timeline is a list ordered by media_PTS. I was wondering how we order the list when it is a mixture of groupID and media_PTS? For example -  if we have two entires for group 67, one with media_PTS and one without - which one do we list first?

Regarding Object ID, lets keep it but
1. If it is omitted, then it is assumed to be 0.
2. If it is unknown, it is signalled as -1. 
Can you add these caveats in a commit?

yekuiwang: Regarding Object ID, sure, I will add these caveats in a commit.

Regarding ordering the list when some entries don't have media_PTS but have group ID, I thought it for a while, but could not figure out a clean way, in particular because the Object ID may also be unknown.

One approach I thought of is to add something like the following: "For any two rows rowA and rowB, wherein rowA has MEDIA_PTS present and equal to mediaPtsA and rowB has MEDIA_PTS absent but the implied media presentation timestamp equal to mediaPtsB, the two rows MUST be ordered in increasing values of mediaPtsA and mediaPtsB." 

However, with this approach, if the entity that creates the rows knows the implied media presentation timestamp of rowB, why wouldn't it add the value into rowB?

Another approach I thought of is to add something like the following: "For any two rows rowA and rowB, wherein rowA has MEDIA_PTS present and equal to mediaPtsA and rowB has MEDIA_PTS absent, if both rows have GROUP_ID present and equal to different values, the two rows are ordered in increasing order of GROUP_ID values, otherwise if both rows have GROUP_ID present and equal to the same value and both have OBJECT_ID specified, the two rows are ordered in increasing of OBJECT_ID values, otherwise the order of the two rows is not important and either one can precede the other one." 

With this approach, it is assumed that any two rows with the same GROUP_ID value and OBJECT_ID specified (either not present thus 0 or present but not -1), the OBJECT_ID values MUST be different. This approach is sort of complicated to specify and I am not sure it is good enough. 

@gwendalsimon what do you think?

My thinking herein is as follows: Unless there is a good way of specifying the ordering, we can mandate the presence of both MEDIA_PTS and GROUP_ID as @wilaw suggested.

yekuiwang: The commit, [Update OBJECT_ID semantics](https://github.com/moq-wg/warp-streaming-format/pull/62/commits/f963022deea4d17189ab9e167a1cd7dae9fd2a16), adds those caveats regarding Object ID.

gwendalsimon: I understand we want to say that the entries of a timeline must be sorted in the order of the media presentation. Any way to compare two entries must be exploited to sort the entries.
Since Locations increase with MEDIA_PTS in WARP, comparing two entries is trivial if
- the two entries have different MEDIA_PTS, or
- the two entries have different Locations (GROUP_ID and OBJECT_ID), or
- the two entries have different GROUPD_IDs,

In the absence of any information enabling comparison between two entries, either one can precede the other one.

Some cases may lead to an error at the client side:
- rowA appears before rowB although MEDIA_PTS(a) is larger than MEDIA_PTS(b)
- rowA appears before rowB although Location(a) is larger than Location(b)
- rowA and rowB have inconsistent MEDIA_PTS and Locations, for example MEDIA_PTS(a)>MEDIA_PTS(b) and Location(a)<Location(b)

If the subscriber notices a sorting error in a Timeline, it MUST terminates the subscription since the Timeline may be corrupted.

wilaw: @gwendalsimon @yekuiwang  - I have been thinking about the issue of accommodating unknown object ID and Group ID in the timeline track. 

Conceptually, the timeline track is produced by the original publisher to describe the history of the content it is producing. This original publisher has intimate and complete knowledge of all the groups and objects that are being produced. Therefore we should mandate that every row of the timeline track MUST include a Media_pts, GroupID, ObjectID and wallclock time, as this information is always known to the original publisher. 

How do we accommodate some of the use cases described in which a 3rd party system is injecting ad breaks or game scores into the track and it doesn't know all this information? 
1. The original publisher has an API which 3rd party systems can use to inject metadata. IN this case, the original publisher is doing the injecting and so it can insert the correct groupID, objectID, media-PTS etc. 
2. 3rd party system (perhaps at smart edge relays) inject metadata, but do it at object boundaries previously established by the original producer. 
3. If a system wants to inject metadata but it can't do one of the above and it doesn't always know the groupID, objectID then IT MUST USE A SEPARATE TRACK. For example, you could make a track of ad breaks that just include wallcock time.

I think this approach keeps the timeline track clean as a record of what has been produced in the past.

yekuiwang: @gwendalsimon mentioned above that "Locations increase with MEDIA_PTS in WARP". That caught my eyes and triggered a thinking on whether we really should mandate the timeline rows are strictly ordered in increasing MEDIA_PTS values. In the context of MOQT, Locations increase with bitstream order (aka decoding order or decode order), not necessarily increase with media presentation time values. Therefore, because of that, shouldn't timeline rows be ordered in increasing of Locations instead? Or at least, a MUST wording should not be used for ordering of the timeline rows based on increasing MEDIA_PTS values?

Regarding @wilaw's use case 3 above, I agree. In a sense, use cases 1 and 2 follow the media timeline, while use case 3 follow a different timeline, which could be called an event timeline. Maybe we name the existing timeline track as media timeline track, and add a definition of event timeline track that just includes wallclock time and some metadata that indicates properties of the event.

wilaw: > Therefore, because of that, shouldn't timeline rows be ordered in increasing of Locations instead? Or at least, a MUST wording should not be used for ordering of the timeline rows based on increasing MEDIA_PTS values?

Yes, this makes sense. Since WARP explicitly indicates that group IDs MUST increase, they are a reliable field on which to order the table. We should move them to the first column. So field order can be
GroupID, ObjectID, Media_PTS, Wallclock , METADATA. @yekuiwang can you make that change as part of this PR?

> Maybe we name the existing timeline track as media timeline track, and add a definition of event timeline track that just includes wallclock time and some metadata that indicates properties of the event.

Strongly agree. Rename "Timeline track" - > "Media Timeline". Create new track type called 'Event Timeline". Since this PR is getting long and crowded, I have opened  a new issue #66  to discuss and track both these changes.

yekuiwang: @wilaw OK, let me try to make changes needed to have the fields ordered as GroupID, ObjectID, Media_PTS, Wallclock, METADATA, as part of this PR. (done, see the commit "[Reorder timeline row elements](https://github.com/moq-wg/warp-streaming-format/pull/62/commits/b4f51114fdf782fe4df79ad116df3a516d91de4e)".)

> Strongly agree. Rename "Timeline track" - > "Media Timeline". Create new track type called 'Event Timeline". Since this PR is getting long and crowded, I have opened a new issue https://github.com/moq-wg/warp-streaming-format/issues/66 to discuss and track both these changes.

Great!

draft-ietf-moq-warp.md
```diff
@@ -945,19 +945,19 @@ Comma-Separated Values (CSV)" Files {{RFC4180}}. The separator is a comma and
...
 MUST be specified as "text/csv" in the catalog.
 
-Each timeline track begins with a header row of MEDIA_PTS,GROUP_ID,OBJECT_ID,
+Each timeline track begins with a header row of GROUP_ID,OBJECT_ID,MEDIA_PTS,
 WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
 
+* GROUP_ID: the MOQT Group ID. This entry MUST not be empty.
+* OBJECT_ID: the MOQT Object ID. This entry MAY be empty. If this entry is
+  present and equal to -1, the Object ID is unknow. Otherwise if this entry
+  is absent, the the Object ID is 0.
```
```suggestion
  present and equal to -1, then Object ID is unknown. Otherwise if this entry
  is absent, then the Object ID is 0.
```

wilaw (approved):

draft-ietf-moq-warp.md
```diff
@@ -419,14 +418,14 @@ assumed to match that of the track declaring the dependencies.
 Location: T    Required: Optional   JSON Type: Number
 
 A number identifying the temporal layer/sub-layer encoding of the track,
-starting with 0 for the base layer, and increasing with higher temporal
-fidelity.
+starting with 0 for the base layer, and increasing by 1 for the next higher
+ temporal fidelity.
```
```suggestion
temporal fidelity.
```

PR #67, opened 2025-09-19
Adding Event timeline track

### LINKED ISSUE #66: Rename timeline track and create new Event Timelines
From https://github.com/moq-wg/warp-streaming-format/pull/62#discussion_r2316675779, suggestion to rename "Timeline" track to "Media Timeline" track as well as introduce a new "Event" Timeline.

The idea is that the media timeline purely describes the historical content produced by the original publisher. 

3rd party events are described in separate tracks. There are two timebases against which we can document events - media time and Unix wallclock time. Wallclock makes sense for live content, however it cannot be used for VOD content. Mixing the two creates a awkward timeline. ThereforeI therefore propose that we create two types of these event tracks:

**Wallclock Events** - used to track events which happen at an explicit wallclock time. This is useful as it allows ingesting events which are decoupled from the media production. Use cases include sports scores, where the game score comes from one service and the audio and video from another. 

**Media Time Event** - used to track events which happen at an explicit media time. Use case would be tracking clip boundaries for goals scored in a live soccer games. Another use case would be SCTE markers coming from the ingest signal. 

For both of these new tracks, instead of an unstructured metadata field, we might consider a more structured syntax. The use of CSV quickly becomes unwieldy for more complex data types. I think we should switch to JSON now. Here's an idea in which the wallclock timeline track is an array of time|type|data objects. 

```
[
    {
        "t": 1756885678361,
        "type": "example.com/custom/live-score-data",
        "data": {
            "status": "in_progress",
            "period": 1,
            "clock": "12:00",
            "homeScore": 0,
            "awayScore": 0,
            "lastPlay": "Game Start"
        }
    },
    {
        "t": 1756885981542,
        "type": "example.com/custom/live-score-data",
        "data": {
            "period": 1,
            "clock": "09:25",
            "homeScore": 2,
            "awayScore": 0,
            "lastPlay": "LAL: #6 LeBron James makes 2-pt jump shot"
        }
    }
]
```

wilaw:
This PR renames the timeline track to "Media timeline", changes its definition so that it refers only to the media published by the original publisher and then adds a new Event timeline track defintion to hold externally generated timeline data. 

Fixes #66

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -957,11 +957,11 @@ in Group IDs using the MOQT Prior Group ID Gap Extension header.
 
 # Media Timeline track {#mediatimelinetrack}
 The media timeline track provides data about the previously published groups and their
-relationship to wallclock time and media time. Media timeline tracks allow players to seek to
-precise points behind the live head in a live broadcast, or for random access in a
-VOD asset. The optional metadata can describe a characteristic of any record in the media 
-timeline. Media timeline tracks are optional. Multiple media timeline tracks can exist inside a
-catalog.
+relationship to wallclock time and media time. Media timeline tracks allow players to
+seek to precise points behind the live head in a live broadcast, or for random access
+in a VOD asset. The optional metadata can describe a characteristic of any record in
+the media timeline. Media timeline tracks are optional. Multiple media timeline tracks
+can exist inside a catalog.
 
 ## Media Timeline track payload {#mediatimelinepayload}
 The payload of a media timeline track is a UTF-8 encoded CSV text file. This
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1033,7 +1033,7 @@ An event timeline track MUST carry a 'type' identifier in the Catalog with a val
 of "eventtimeline". An event timeline track MUST carry a 'depends' attribute which
 contains an array of all track names to which the event timeline track applies.
 
-## Event Timeline track updating.
+## Event Timeline track updating
 The publisher MUST publish an indepdendent event timeline in the first MOQT Object
 of each MOQT Group of an event timeline track. The publisher MAY publish incremental
 updates in the second and subsequent Objects within each Group. Incremental updates
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -334,10 +334,11 @@ as defined in Table 3.
 
 Table 3: Allowed packaging values
 
-| Name            |   Value   |      Reference        |
-|:================|:==========|:======================|
-| LOC             | loc       | See RFC XXXX          |
-| Timeline        | timeline  | See {{timelinetrack}} |
+| Name            |   Value        |      Reference             |
+|:================|:===============|:===========================|
+| LOC             | loc            | See RFC XXXX               |
+| Media Timeline  | mediatimeline  | See {{mediatimelinetrack}} |
+| Event Timeline  | eventtimeline  | See {{eventtimelinetrack}} |
 
 ### Track role {#trackrole}
 Location: T    Required: Optional   JSON Type: String
@@ -355,7 +356,8 @@ Table 4: Reserved track roles
 | audiodescription | An audio description for visually impaired users           |
 | video            | Visual content                                             |
 | audio            | Audio content                                              |
-| timeline         | A WARP timeline {{timelinetrack}}                          |
+| mediatimeline    | A WARP media timeline {{mediatimelinetrack}}               |
+| eventtimeline    | A WARP event timeline {{eventtimelinetrack}}               |
 | caption          | A textual representation of the audio track                |
 | subtitle         | A transcription of the spoken dialogue                     |
 | signlanguage     | A visual track for hearing impaired users.                 |
@@ -953,23 +955,22 @@ Each subsequent Group ID MUST increase by 1.
 If a publisher is able to maintain state across a republish, it MUST signal the gap
 in Group IDs using the MOQT Prior Group ID Gap Extension header.
 
-# Timeline track {#timelinetrack}
-The timeline track provides data about the previously published groups and their
-relationship to wallclock time, media time and associated timed-metadata.
-Timeline tracks allow players to seek to precise points behind the live head in
-a live broadcast, or for random access in a VOD asset. A timeline track might also
-be used to insert events at media times which do not correlate with Object
-boundaries. Timeline tracks are optional. Multiple timeline tracks can exist
-inside a catalog.
+# Media Timeline track {#mediatimelinetrack}
+The media timeline track provides data about the previously published groups and their
+relationship to wallclock time and media time. Timeline tracks allow players to seek to
+precise points behind the live head in a live broadcast, or for random access in a
+VOD asset. The optional metadata can describe a characteristic of any record in the
+timeline. Timeline tracks are optional. Multiple timeline tracks can exist inside a
+catalog.
 
-## Timeline track payload
-The payload of a timeline track is a UTF-8 encoded CSV text file. This payload
-is formatted according to RFC4180 "Common Format and MIME Type for
+## Media Timeline track payload {#mediatimelinepayload}
+The payload of a media timeline track is a UTF-8 encoded CSV text file. This
+payload is formatted according to RFC4180 "Common Format and MIME Type for
 Comma-Separated Values (CSV)" Files {{RFC4180}}. The separator is a comma and
-each line is separated by a carriage return. The mime-type of a timeline track
-MUST be specified as "text/csv" in the catalog.
+each line is separated by a carriage return. The mime-type of a media timeline
+track MUST be specified as "text/csv" in the catalog.
 
-Each timeline track begins with a header row of GROUP_ID,OBJECT_ID,MEDIA_PTS,
+Each media timeline track begins with a header row of GROUP_ID, OBJECT_ID, MEDIA_PTS,
 WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
 
 * GROUP_ID: the MOQT Group ID. This entry MUST not be empty.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-warp.md
```diff
@@ -953,23 +955,22 @@ Each subsequent Group ID MUST increase by 1.
...
+The payload of a media timeline track is a UTF-8 encoded CSV text file. This
+payload is formatted according to RFC4180 "Common Format and MIME Type for
 Comma-Separated Values (CSV)" Files {{RFC4180}}. The separator is a comma and
-each line is separated by a carriage return. The mime-type of a timeline track
-MUST be specified as "text/csv" in the catalog.
+each line is separated by a carriage return. The mime-type of a media timeline
+track MUST be specified as "text/csv" in the catalog.
 
-Each timeline track begins with a header row of GROUP_ID,OBJECT_ID,MEDIA_PTS,
+Each media timeline track begins with a header row of GROUP_ID, OBJECT_ID, MEDIA_PTS,
```
gwendalsimon, line 974: Location is now formally defined in MOQT spec https://www.ietf.org/archive/id/draft-ietf-moq-transport-14.html#section-1.4.1 so, instead of having GROUP_ID and OBJECT_ID in the timeline, we could prefer LOCATION. If the formatting of Location changes in the future, the media timeline would stay valid.

draft-ietf-moq-warp.md
```diff
@@ -953,23 +955,22 @@ Each subsequent Group ID MUST increase by 1.
...
-inside a catalog.
-
-## Timeline track payload
-The payload of a timeline track is a UTF-8 encoded CSV text file. This payload
-is formatted according to RFC4180 "Common Format and MIME Type for
+# Media Timeline track {#mediatimelinetrack}
+The media timeline track provides data about the previously published groups and their
+relationship to wallclock time and media time. Timeline tracks allow players to seek to
+precise points behind the live head in a live broadcast, or for random access in a
+VOD asset. The optional metadata can describe a characteristic of any record in the
```
gwendalsimon: Should we keep a metadata field in the Media Timeline? 
In my understanding, the media timeline provides all the info for seeking while any additional information should be given in the event timeline.
Do you have some examples of metadata that could be necessary to include in the media timeline?

draft-ietf-moq-warp.md
```diff
@@ -991,20 +992,91 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+updates in the second and subsequent Objects within each GROUP. Incremental updates
+only contain media timeline records since the last media timeline Object.
+
+# Event Timeline track {#eventtimelinetrack}
+The event timeline track provides a mechanism to associate ad-hoc event metadata with
+the broadcast. Use-case examples include live sports score data, GPS coordinates of race
+cars, or active speaker notifications in web conferences.
+
+To allow the client to bind this event metadata with the broadcast content described by
+the media timeline track, each event record MUST contain a reference to either GroupID,
```
gwendalsimon: Location instead of GroupID?

gwendalsimon: (english writing) Can we say "either" when we have three element? It is vague.

We should prefer:
- only one of three identifiers (Location, Media_PTS, Wallclock time)
- any of the three identifiers  ==> In that case, the formatting may become a problem and we may need to write some sentences about possible inconsistencies.

I would be in favor of having only one of the three.

draft-ietf-moq-warp.md
```diff
@@ -991,20 +992,91 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+Event timeline tracks are optional. Multiple event timeline tracks can exist inside a
+catalog.
+
+## Event Timeline data format {#eventtimelineformat}
+An event timeline track is a JSON {{JSON}} document. It contains an array of records.
+Each record consists of a JSON Object containing the following required fields:
+
+* A timeline reference, which MUST be either 't' for wallclock time, 'g' for Group ID or
+  'm' for Media PTS. Typically only one timeline reference is included in each record,
+  although multiple timeline entries may exist within a record, as long as they
```
gwendalsimon: Yes, I can see my comment https://github.com/moq-wg/warp-streaming-format/pull/67/files#r2362812236 is addressed here. So you may ignore. I keep it in case we want to discuss it.

gwendalsimon (commented):
Thank you for the PR. I added minor comments, but overall it is good to me.

draft-ietf-moq-warp.md
wilaw: Yes, I thought we could remove it too, however @yekuiwang  had a use-case, of describing the type of SAP access point represented by the Group. So I think there is some valid metadata that can describe the produced content. In many cases, it may be empty.

draft-ietf-moq-warp.md
wilaw: Location defines a binary serialization. We have a JSON doc here. Do you propose somehting like this? 

> [
>     {
>         "location": { "g":432, "o": 3},
>         "type": "com.example.custom.live-score-data",
>         "data": {
> ...

or can we do an array

> [
>     {
>         "location": [432, 3],
>         "type": "com.example.custom.live-score-data",
>         "data": {
> ...

Note that location doesn't allow the Object ID to be undefined or have a default value, so if its used then both groupID and object ID must be supplied.

wilaw: I'm fine with one. Simplicity is  always the better choice.

wilaw: Commit https://github.com/moq-wg/warp-streaming-format/pull/67/commits/157d5d655fdf6bc2be1a2c5e5354bf0b5fa95c9a restricts the record to only supplying one identifier.

draft-ietf-moq-warp.md
```diff
@@ -953,23 +955,22 @@ Each subsequent Group ID MUST increase by 1.
...
-## Timeline track payload
-The payload of a timeline track is a UTF-8 encoded CSV text file. This payload
-is formatted according to RFC4180 "Common Format and MIME Type for
+# Media Timeline track {#mediatimelinetrack}
+The media timeline track provides data about the previously published groups and their
+relationship to wallclock time and media time. Timeline tracks allow players to seek to
+precise points behind the live head in a live broadcast, or for random access in a
+VOD asset. The optional metadata can describe a characteristic of any record in the
+timeline. Timeline tracks are optional. Multiple timeline tracks can exist inside a
+catalog.
```
yekuiwang:
```suggestion
relationship to wallclock time and media time. Media timeline tracks allow players to seek to
precise points behind the live head in a live broadcast, or for random access in a
VOD asset. The optional metadata can describe a characteristic of any record in the media 
timeline. Media timeline tracks are optional. Multiple media timeline tracks can exist inside a
catalog.

File: draft-ietf-moq-warp.md [hunk 6]
```diff
@@ -991,20 +992,90 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
-## Timeline Catalog requirements
-A timeline track MUST carry a 'type' identifier in the Catalog with a value of
-"timeline". A timeline track MUST carry a 'dependes' attribute which
-contains an array of all track names to which the timeline track applies.
+## Media Timeline Catalog requirements
+A media timeline track MUST carry a 'type' identifier in the Catalog with a value
+of "mediatimeline". A media timeline track MUST carry a 'depends' attribute which
+contains an array of all track names to which the media timeline track applies.
+
+## Media Timeline track updating.
```
yekuiwang:
```suggestion
## Media Timeline track updating

File: draft-ietf-moq-warp.md [hunk 7]
```diff
@@ -991,20 +992,90 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+The publisher MUST publish an indepdendent media timeline in the first MOQT Object
+of each MOQT Group of a media timeline track. The publisher MAY publish incremental
+updates in the second and subsequent Objects within each GROUP. Incremental updates
```
yekuiwang:
```suggestion
updates in the second and subsequent Objects within each Group. Incremental updates

File: draft-ietf-moq-warp.md [hunk 8]
```diff
@@ -991,20 +992,90 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+only contain media timeline records since the last media timeline Object.
+
+# Event Timeline track {#eventtimelinetrack}
+The event timeline track provides a mechanism to associate ad-hoc event metadata with
+the broadcast. Use-case examples include live sports score data, GPS coordinates of race
+cars, or active speaker notifications in web conferences.
+
+To allow the client to bind this event metadata with the broadcast content described by
+the media timeline track, each event record MUST contain a reference to either GroupID,
+Media PTS or wallclock time.
```
yekuiwang:
```suggestion
the media timeline track, each event record MUST contain a reference to GroupID,
Media PTS or wallclock time.

File: draft-ietf-moq-warp.md [hunk 9]
```diff
@@ -991,20 +992,90 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+
+## Event Timeline Catalog requirements
+An event timeline track MUST carry a 'type' identifier in the Catalog with a value
+of "eventtimeline". An event timeline track MUST carry a 'depends' attribute which
+contains an array of all track names to which the event timeline track applies.
+
+## Event Timeline track updating.
+The publisher MUST publish an indepdendent event timeline in the first MOQT Object
+of each MOQT Group of an event timeline track. The publisher MAY publish incremental
+updates in the second and subsequent Objects within each GROUP. Incremental updates
```
yekuiwang:
```suggestion
updates in the second and subsequent Objects within each Group. Incremental updates

**Comment by yekuiwang** [comment]
@wilaw BTW, we are going to use JSON also for media timeline later on, right?

**Comment by wilaw** [comment]
> @wilaw BTW, we are going to use JSON also for media timeline later on, right?

Yes. But I'll make that a separate PR instead of overloading this one. 

I opened https://github.com/moq-wg/warp-streaming-format/issues/68 to track that.

File: draft-ietf-moq-warp.md [hunk 10]
```diff
@@ -991,20 +992,90 @@ WALLCLOCK,METADATA. This row defines the 5 columns of data within each record.
...
+* A 'type' field, which is a String defining the type of data contained within the data
+  field. Types are defined by the application provider.
+* A 'data' Object, whose structure is defined by the 'type' value.
+
+## Event Timeline Catalog requirements
+An event timeline track MUST carry a 'type' identifier in the Catalog with a value
+of "eventtimeline". An event timeline track MUST carry a 'depends' attribute which
+contains an array of all track names to which the event timeline track applies.
+
+## Event Timeline track updating.
```
yekuiwang:
```suggestion
## Event Timeline track updating

yekuiwang (approved):

draft-ietf-moq-warp.md
wilaw: Commit https://github.com/moq-wg/warp-streaming-format/pull/67/commits/d4ad651876505e5ffe2e797da6572977f81a04b4 switches to using Location

draft-ietf-moq-warp.md
wilaw: Commit https://github.com/moq-wg/warp-streaming-format/pull/67/commits/d4ad651876505e5ffe2e797da6572977f81a04b4 defines the JSON representation of Location using an array of numbers.

PR #70, opened 2025-10-08
Changing media timeline to JSON

### LINKED ISSUE #68: Convert timeline track to JSON format
We are using JSON for the catalog and the Event Timeline, so the request is to convert Media Timeline from a CSV to a JSON format.

Benefits
1. Consist formatting allows a simple parser in the player
2. JSON is cleaner for sparse data (such as SAP types) and occasional metadata.
3. JSON more flexible and not limited to fixed columnar data.

Cons:
1. JSON more verbose than CSV for same payload. However we can zip the JSON plus if we really care about compression we should shift to a binary format.

wilaw:
This PR converts the media timeline track to a JSON format.

It also modifies the Event timeline track to externalize the "type" to the catalog, so that a client can choose an event timeline based on the type of data it contains without having to load it first to discover that type. 

Fixes #68

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1105,11 +1105,11 @@ a JSON Object containing the following required fields:
 ## Event Timeline Catalog requirements
 An event timeline track MUST carry:
 
-* a {{packagingtype}} attribute with a value of "eventtimeline".
-* a {{depends}} attribute which contains an array of all track names to which the event
+* a {{packaging}} attribute with a value of "eventtimeline".
+* a {{dependencies}} attribute which contains an array of all track names to which the event
   timeline track applies.
 * a {{mimetype}} attribute with a value of "application/json".
-* an {{eventtimetype}} attribute declaring the type & structure of data contained in the
+* an {{eventtype}} attribute declaring the type & structure of data contained in the
   event timeline track.
 
 ## Event Timeline track updating
@@ -1181,7 +1181,7 @@ publisher can deliver a deterministic signal to all subscribers that the broadca
 is complete by taking the following steps:
 
 * Send a SUBSCRIBE_DONE (See MOQT Sect 8.1.2) message for all active tracks using
-  status code 0x2	Track Ended.
+  status code 0x2 Track Ended.
 * If the live stream is being converted instantly to a VOD asset, then publish an
   independent (non-delta) catalog update which, for each track, sets isLive {{islive}}
   to FALSE and adds a track duration {{trackduration}} field.
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -218,6 +218,7 @@ Table 1 provides an overview of all fields defined by this document.
 | Track namespace         | namespace              | {{tracknamespace}}        |
 | Track name              | name                   | {{trackname}}             |
 | Packaging               | packaging              | {{packaging}}             |
+| Event timeline type     | eventType              | {{eventtype}              |
 | Is Live                 | isLive                 | {{islive}}                |
 | Target latency          | targetLatency          | {{targetlatency}}         |
 | Track role              | role                   | {{trackrole}}             |
@@ -340,6 +341,17 @@ Table 3: Allowed packaging values
 | LOC             | loc            | See RFC XXXX               |
 | Media Timeline  | mediatimeline  | See {{mediatimelinetrack}} |
 | Event Timeline  | eventtimeline  | See {{eventtimelinetrack}} |
+
+
+### Event timeline type {#eventtype}
+Location: T    Required: Optional   JSON Type: String
+
+A String defining the type & structure of the data contained within the data
+field of the Event timeline track. Types are defined by the application provider
+and are not centrally registered. Implementers are encouraged to use a unique
+naming scheme, such as Reverse Domain Name Notation, to avoid naming collisions.
+This field is required if the {{packaging}} value is "eventtimeline".
+This field MUST NOT be used if the packaging value is not "eventtimeline". 
 
 ### Track role {#trackrole}
 Location: T    Required: Optional   JSON Type: String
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-warp.md
```diff
@@ -920,6 +932,65 @@ and video tracks.
...
+      "mimetype": "application/json",
+      "depends": ["1080p-video","audio"],
+    },
+    {
+      "name": "identified-objects",
+      "namespace": "another-provider/time-synchronized-data",
+      "packaging": "eventtimeline",
+      "eventTimelineType": "com.ai-extraction/appID/v3"
+      "mimetype": "application/json",
+      "depends": ["1080p-video"],
```
gwendalsimon, line 960: We could add one line to express which index is used by the event timeline (out of the three elements of the media timeline). For example "index": "mediaPTS",

gwendalsimon (approved):
I just have a remark about the type used for the index. It is really minor.

draft-ietf-moq-warp.md
```diff
@@ -960,43 +1031,42 @@ in Group IDs using the MOQT Prior Group ID Gap Extension header.
...
-* MEDIA_PTS: a media timestamp rounded to the nearest millisecond. This entry
-  MUST NOT be empty. If the Object ID entry is present and equal to -1, this
-  value MUST match the media presentation timestamp of the first media sample
-  in an Object that is in the Group but is not the first Object in the Group.
-  Otherwise, this value MUST match the media presentation timestamp of the
-  first media sample in the referenced Object.
-* WALLCLOCK: the wallclock time at which the media was encoded, expressed as
+A media timeline track is a JSON {{JSON}} document. This document MAY be compressed
+using GZIP {{GZIP}}. The document contains an array of records. Each record consists of
+an array of three required items, whose ordinal position defines their type:
```
yekuiwang, line 1040: This is extendable, to have additional, optional items, right?

draft-ietf-moq-warp.md
```diff
@@ -960,43 +1031,42 @@ in Group IDs using the MOQT Prior Group ID Gap Extension header.
...
+
+* The first item holds the media presentation timestamp, expressed as a JSON Number.
+  This value MUST match the media presentation timestamp, rounded to the nearest
+  millisecond, of the first media sample in the referenced Object
+* The second item holds the MOQT Location of the entry, defined as a tuple of the MOQT
+  Group ID and MOQT Object ID, and expressed as a JSON Array of Numbers, where the
+  first number is the Group ID and the second number is the Object ID.
```
yekuiwang, line 1047: We established earlier that the Object ID may be unknown. We also allowed the Object ID to be absent and defaulted to 0. Are these possible in the JSON format? If possible, I think we should keep these.

draft-ietf-moq-warp.md
gwendalsimon: In this writing, the Media Timeline is not extendable. It contains these three items. Any additional item related to an line must be in a separate Event Timeline track. 
In essence, the goal of the Media Timeline is to make the link between PTS, Locations, and Time. Any other metadata is by definition optional, and so should be in a separate track.
Unless you have a use case in mind?

draft-ietf-moq-warp.md
gwendalsimon: The Media Timeline being generated by the Publisher, the Object ID cannot be unknown. 
In Event timeline, it might be possible that the publisher of the event timeline does not know the object ID. By default, I guess we can mandate to be set to Object ID = 0. Though, do you have a use-case where the publisher of an event timeline does not know neither PTS, nor Wallclock time, nor Object ID. It would be quite imprecise, wouldn't it?

gwendalsimon: To give a bit more context, I think that we should keep "Location" as the type of items in Media Timeline because it is a well-defined abstract concept in MOQT, and it may be possible that, in the future, MOQT Location definition changes. We need to stay consistent.

draft-ietf-moq-warp.md
yekuiwang: In the latest [CARP draft](https://wilaw.github.io/carp/draft-law-moq-carp.html), [Section 3.6](https://wilaw.github.io/carp/draft-law-moq-carp.html#name-timeline-description),  we have extensions to the Media Timeline through metadata, with concrete extensions for signalling of [SAP type](https://wilaw.github.io/carp/draft-law-moq-carp.html#name-metadata-signalling-of-sap-) and [earliest presentation time](https://wilaw.github.io/carp/draft-law-moq-carp.html#name-metadata-signalling-of-earl). If we get rid of metadata and at the same time make it non-extensible, then herein we are not only changing the format to JSON, but also we are losing capabilities we had earlier for timeline track.

draft-ietf-moq-warp.md
yekuiwang: > The Media Timeline being generated by the Publisher, the Object ID cannot be unknown. In Event timeline, it might be possible that the publisher of the event timeline does not know the object ID. By default, I guess we can mandate to be set to Object ID = 0. Though, do you have a use-case where the publisher of an event timeline does not know neither PTS, nor Wallclock time, nor Object ID. It would be quite imprecise, wouldn't it?

During earlier discussions of media timeline, Will mentioned that it is possible that the Object ID may be unknown. Will?

wilaw: HI @yekuiwang   -yes, it was true that when we had a single timeline track, which carried media timeline data mixed in with every possible type of arbitrary 3rd party timeline data, that there were cases where the 3rd party data would not know the object ID. Hence we invented a signaling scheme with special meaning for Objects IDs that were missing vs unknown. We also need special cases for GroupIDs that were missing or unknown and pretty quickly we were working our way into a hacky corner of brittle conventions. 

To address that, at the Toronto interim, I presented the concept that we can have multiple timeline tracks - one of which, the media timeline, is purely dedicated to describing the output of the original publisher in a strict tuple of media time, Locaiton (Group,Object) and Wallclock time, all of these being known to the publisher. We then can have multiple other event timeline tracks which carry information that is synched to _one_ of the primary indices (media time, location or wallclock). This is where information such on such as SAP type and earliest presentation time, can be included. This concept was approved by the meeting and hence I made this PR as well as https://github.com/moq-wg/warp-streaming-format/pull/67 to make those changes.

draft-ietf-moq-warp.md
wilaw: @yekuiwang  - see https://github.com/moq-wg/warp-streaming-format/pull/70#discussion_r2416061804. Don't worry - I will prepare a PR against CARP to show how to transmit SAP type and EPT using an event timeline instead of the media timeline.

draft-ietf-moq-warp.md
yekuiwang: Aha, during #67 discussions, e.g., [herein](https://github.com/moq-wg/warp-streaming-format/pull/67#discussion_r2363301228), we were still saying to keep metadata in the media timeline, and I was not aware that later on there was a change of a mind on this as part of the Toronto discussion. But fine, if the intent is to have information such as SAP type and earliest presentation time conveyed through event timeline tracks, although these two pieces of information really are part of the media properties not really events.

draft-ietf-moq-warp.md
wilaw: See https://github.com/wilaw/carp/pull/11

PR #72, opened 2025-10-31
Defining a connection URL syntax and method of connection

### LINKED ISSUE #60: URL syntax identifying both connection and track
It would be convenient to be able to represent both a connection to a MOQT endpoint and a WARP catalog track as a URL. This would allow a Content Management System to vend a single URL to a player which would give it the information it needs to both connect to a MOQT distribution network (MDN) and discover a catalog track.

WebTransport endpoints have the form of a http3 connection and may have several path segments for business purposes

`https://mdn.com/customerID/appID`

A track can be represented as a namespace + name as a pseudo relative URL

`example.com/live/broadcastID/catalog`

We can combine these and add in an access token

`https://mdn.com/customerID/appID/example.com/live/broadcastID/catalog?token=12345`

The MDN will know which segments its needs to parse to accept the connection. However, how does the player know where the namespace begins in order to issue the SUBSCRIBE? Some options:

1. The player knows through some out-of-band information and we don't need to tell it. This seems fragile for interop. 

2. We specify that the first segment after the hostname is always reserved for the connection.  Separators other than / may be used by the MDN vendor to provide ancillary information.

`https://mdn.com/customerID-appID/example.com/live/broadcastID/catalog?token=12345`

3. We create a reserved path segment which identified the division between the two. In this case we use '!' 

`https://mdn.com/customerID/appID/!/example.com/live/broadcastID/catalog?token=12345`

A second problem is that we need to tell the player that the query arg holds a token which should be extracted and supplied via the AUTHORIZATION TOKEN setup parameter , as well as a token for SUBSCRIBE, PUBLISH etc. For this, we can reserve a query arg name , for example 'warp-token'. Providers  are not required to supply a token in this fashion (they may vend it as a separate asset to the player), but at least it provides an interoperable methd of providing a token to players which the content distributor does not control. 

`https://mdn.com/customerID/appID/!/example.com/live/broadcastID/catalog?warptoken=12345`

wilaw:
Defines a URL syntax along with methods of connecting over WebTransport and raw QUIC. 

Example URLs
* URL pointing at a catalog
`  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID`

* URL pointing at a non-catalog track
 ` moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&t=video`

* URL pointing at a subclip of a catalog
`  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16`

* URL pointing at a catalog and supplying a token
```
  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
  zMwNDM4NDAw
```

Fixes #60

### DIFF
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1229,7 +1229,7 @@ content over the range specified.
 If 't' is omitted, then the URL is assumed to point at a catalog, in which case the track
 name is automatically 'catalog'.
 If 'ns' and 't' are omitted, then the player is assumed to have out-of-band information
-for accessing tracks and the URL purely defines the connection to the delivery network. 
+for accessing tracks and the URL purely defines the connection to the delivery network.
 
 Example query args
 * wallclock-range=1761759637565-1761759836189
@@ -1250,7 +1250,7 @@ Example URLs
   moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
   lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
   zMwNDM4NDAw
- 
+
 ### Connection using WebTransport
 Assuming a WARP URL of
 
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1254,9 +1254,9 @@ Example URLs
 ### Connection using WebTransport
 Assuming a WARP URL of
 
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
-  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
-  zMwNDM4NDAw
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
+  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
+  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
 
 1. The player generates the WebTransport URL by substituting https for moqt and stripping off
    all query args. The token bytes are extracted and stored separately.
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1177,7 +1177,7 @@ A WARP URL is a String with the following components
 
 moqt:// + authority + path + query
 
-* THe URL MUST conform with {{RFC3986}}.
+* The URL MUST conform with {{RFC3986}}.
 * The protocol is moqt and is required.
 * The authority is required and holds the host and optional port. The default port
   will be 443. The authority holds the information needed for the client to connect
@@ -1247,16 +1247,16 @@ Example URLs
   moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16
 
 * URL pointing at a catalog and supplying a token
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
-  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
-  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
+  zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
+  NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
 
 ### Connection using WebTransport
 Assuming a WARP URL of
 
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
-  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
-  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
+moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
+zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
+NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
 
 1. The player generates the WebTransport URL by substituting https for moqt and stripping off
    all query args. The token bytes are extracted and stored separately.
@@ -1269,9 +1269,9 @@ Assuming a WARP URL of
 ### Connection using raw QUIC
 Assuming a WARP URL of
 
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
-  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
-  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
+moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
+zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
+NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
 
 1. The player generates the AUTHORITY value of 'example.com' by stripping off the protocol,
    path and query components from the WARp URL. The token bytes are extracted and stored separately.
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1247,16 +1247,16 @@ Example URLs
   moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16
 
 * URL pointing at a catalog and supplying a token
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
-  zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
-  NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID
+  &c4m=gqhkYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5u
+  ZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM
 
 ### Connection using WebTransport
 Assuming a WARP URL of
 
-moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
-zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
-NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
+moqt://example.com/relay-app/relayID?ns=customerID/broadcastID
+&c4m=gqhkYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5u
+ZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM
 
 1. The player generates the WebTransport URL by substituting https for moqt and stripping off
    all query args. The token bytes are extracted and stored separately.
@@ -1269,9 +1269,9 @@ NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
 ### Connection using raw QUIC
 Assuming a WARP URL of
 
-moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGV
-zaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2
-NzgwMHZpc3VlZF9hdD0xNzMwNDM4NDAw
+moqt://example.com/relay-app/relayID?ns=customerID/broadcastID
+&c4m=gqhkYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5u
+ZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM
 
 1. The player generates the AUTHORITY value of 'example.com' by stripping off the protocol,
    path and query components from the WARp URL. The token bytes are extracted and stored separately.
--- a/draft-ietf-moq-warp.md
+++ b/draft-ietf-moq-warp.md
@@ -1247,9 +1247,9 @@ Example URLs
   moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16
 
 * URL pointing at a catalog and supplying a token
-  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
-  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
-  zMwNDM4NDAw
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
+  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
+  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
 
 ### Connection using WebTransport
 Assuming a WARP URL of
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+* URL pointing at a non-catalog track
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&t=video
+
+* URL pointing at a subclip of a catalog
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16
+
+* URL pointing at a catalog and supplying a token
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
+  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
+  zMwNDM4NDAw
```
```suggestion
  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
```

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
+  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
+  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
+
+### Connection using WebTransport
+Assuming a WARP URL of
+
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
+  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
+  zMwNDM4NDAw
```
```suggestion
  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
```

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+   parameter and registering a token alias.
+3. Player subscribes to NAMESPACE customerID | broadcastID (where | represents the tuple
+   boundary) and to NAME catalog, along with the token alias.
+
+### Connection using raw QUIC
+Assuming a WARP URL of
+
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0eXB
+  lY2NhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xN
+  zMwNDM4NDAw
```
```suggestion
  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&c4m=gqhkYWxnIGVzaGFyqGR0
  eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2Nzgw
  MHZpc3VlZF9hdD0xNzMwNDM4NDAw
```

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+* The authority is required and holds the host and optional port. The default port
+  will be 443. The authority holds the information needed for the client to connect
+  to the distribution service, using either WebTransport or Raw QUIC.
+* The path is optional. If present, it contains information that might be used by
+  the server in initializing and configuring the connection.
+* The query is optional. If present, it holds key-value data. Certain keys are reserved
+  by this specificaiton to hold special data. Multiple key/value pairs MAY be included,
+  separated by an ampersand. Query options MUST be stripped from the path when the path
+  is sent to the server at conneciton establishment. Query arguments are intended only
+  for the player and MUST NOT be transmitted to the server.
```
michalhosna, line 1191: As I read it, not sending query is incompatible with current MOQT wording, which defines `moqt://` handling:

https://www.ietf.org/archive/id/draft-ietf-moq-transport-15.html#section-3.1.2

> The client can establish a connection to an MOQT server identified by a given URI by setting up a QUIC connection to the host and port identified by the authority section of the URI. The authority, path-abempty and **query portions of the URI are also transmitted in SETUP parameters** (see [Section 9.3.1](https://www.ietf.org/archive/id/draft-ietf-moq-transport-15.html#setup-params)).

Switching to a custom scheme sidesteps this problem. .

michalhosna, line 1191: I think only reserved params should be stripped.

As MOQT defines "query," I think it is expected that query can be used to send data to the server, and I do think that we need to keep in some kind of custom extensibility similar to HTTP given that we are trying to replace HTTP in some contexts. 

I do understand that it is not elegant.

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+| t               | The Name of a non-catalog track                  |
+| wallclock-range | A subclip defined by a wallclock time range      |
+| mediatime-range | A subclip defined by a media time range          |
+| location-range  | A subclip defined by a MOQT Location range       |
+| c4m             | A base64 encoded C4M token                       |
+
+ * ns - the Namespace of the track. The '/' character in the String defines
+   tuple field boundaries and is not included in the tuple fields. The '/' character
+   MUST NOT be used other than to define tuple boundaries. A closing '/' for the
+   last tuple field MUST NOT be included. This key MUST be included if 't' is present.
```
michalhosna, line 1215: What about leading `/`? 

I am not sure if I am missing something, but neither MOQT nor WARP seem to disallow zero-sized components. So `//a//` could be ("", "", "a", "", "")?

EDIT: https://github.com/moq-wg/moq-transport/issues/1041 talks about zero length fields

michalhosna, line 1215: I do think we need to describe how to escape `/` if needed.

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+ * t - the track name of a non-catalog track. This field MUST NOT be included if the
+   URL is referencing a catalog track.
```
michalhosna, line 1217: Is `/` allowed in t? Should it be always treated as literal here? I think it should be noted explicitly here that the parser for `t` must be diffent than for `ns`

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+   expressed as Start Group ID, Start Object ID, End Group ID, End Object ID, each
+   separated by a "-" dash.
+
+Only one of wallclock-range, mediatime-range or location-range MAY be included in each
+URL. Inclusion of a range query-arg is an instruction to the player to only play the track
+content over the range specified.
+If 't' is omitted, then the URL is assumed to point at a catalog, in which case the track
+name is automatically 'catalog'.
+If 'ns' and 't' are omitted, then the player is assumed to have out-of-band information
+for accessing tracks and the URL purely defines the connection to the delivery network.
```
michalhosna, line 1232: What if only `t` is present?

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+   last tuple field MUST NOT be included. This key MUST be included if 't' is present.
+ * t - the track name of a non-catalog track. This field MUST NOT be included if the
+   URL is referencing a catalog track.
+ * wallclock-range - a range defined by start and end wallclock times, each expressed
+   as milliseconds since Unix Epoch and separated by a "-" dash.
+ * mediatime-range - a range defined by start and end media times, each expressed
+    as milliseconds and separated by a "-" dash.
+ * location-range - a range defined by start and end media MOQT Location tuples and
+   expressed as Start Group ID, Start Object ID, End Group ID, End Object ID, each
+   separated by a "-" dash.
```
michalhosna, line 1224: Don't we want to support open ranges?

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
 
 # Workflow
 
+## URL construction & interpretation
+A WARP URL is a String with the following components
+
+moqt:// + authority + path + query
+
+* The URL MUST conform with {{RFC3986}}.
```
michalhosna, line 1180: I think there needs to be commentary on percent encoding. 

E.g. May I encode `/` as a tuple separator in the `ns` query arg? Or is that effectively escaping, and `/` must always be non encoded?  

I think RFC3986 is insufficient for this use case, especially given it specifically mentions `/` in query section.

draft-ietf-moq-warp.md
michalhosna, line 1224: How does one use this range for non-catalog tracks?

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+ * ns - the Namespace of the track. The '/' character in the String defines
+   tuple field boundaries and is not included in the tuple fields. The '/' character
+   MUST NOT be used other than to define tuple boundaries. A closing '/' for the
+   last tuple field MUST NOT be included. This key MUST be included if 't' is present.
+ * t - the track name of a non-catalog track. This field MUST NOT be included if the
+   URL is referencing a catalog track.
+ * wallclock-range - a range defined by start and end wallclock times, each expressed
+   as milliseconds since Unix Epoch and separated by a "-" dash.
+ * mediatime-range - a range defined by start and end media times, each expressed
+    as milliseconds and separated by a "-" dash.
```
michalhosna, line 1221: I think wording from [7.1. ][Media Timeline track payload](https://moq-wg.github.io/warp-streaming-format/draft-ietf-moq-warp.html#name-media-timeline-track-payloa) may be taken here.

"media presentation timestamp, rounded to the nearest millisecond"

Sidenote, why are we not using timescale, if catalog defines it? Now I am confusing myself. Does even LOC have PTS?

michalhosna (changes requested):
I apologize for the longish, nitpicky review, but I do think we need more clarity. 

I am happy to help with integrating the commentary; I'm not sure what the best way there. Should I send my own counter-PR?

I think the story around the fragment was not really resolved, and I still think it's more elegant than this. But if we can get better consensus on the query, let's go this way.

vasilvv:
I am confused, is the URL intentionally does not specify the protocol that is used for connecting to the server?

Also, I am still not convinced that using query string is better than using the fragment.

wilaw:
> I am confused, is the URL intentionally does not specify the protocol that is used for connecting to the server?

Correct. The desire was for this to be at the determination of the client. Originally we had proposed a https://.. URL to indicate a Webtransport connection and moqt://... to indicate raw QUIC. However at IETF #123 there was [pushback ](https://github.com/moq-wg/warp-streaming-format/issues/60#issuecomment-3335631745)on this as the https:// URL could be assumed by many systems to have a HTTP GET response. See also [this ](https://github.com/moq-wg/warp-streaming-format/issues/60#issuecomment-3356019486)comment. 

> Also, I am still not convinced that using query string is better than using the fragment.

At the Toronto interim I did propose your fragment concept for the URL - see https://docs.google.com/presentation/d/1pFZ7kdp1s1CY2K6YgRQ1M_9tpwNIWK6j1gNuJj3UoJ0/edit?slide=id.g3630ee1ef9b_0_22#slide=id.g3630ee1ef9b_0_22. The [concern](https://github.com/moq-wg/warp-streaming-format/issues/60#issuecomment-3359672623) was that a fragment response has a Mime-type associated with it and that the fragment construct of pointing at a subset of a resource was being abused to code in additional information not related to the resource. For this, flexible query arguments were considered a cleaner and more extensible fit. It was that feedback in Toronto which drove the establishment of this PR.  

An additional issue with fragment is that there can be only one per URL. For MSF, we need to give the client both a namespace and a name (as they are applied separately in all MOQT messages). So we then need to create formatting rules for the fragment to separate namespace from name. At that point, you are inventing something that looks like KV pairs, so why not just use the well-understood scheme of query args?

vasilvv:
> Correct. The desire was for this to be at the determination of the client.

I don't understand the circumstances under which this would be useful.  Usually I would expect a MOQT resource to be available either via WT or raw QUIC, but not both.  A URL "provide a means of locating the resource by describing its primary access mechanism" (RFC 3986), so not knowing the protocol sounds like a glaring omission.

> The https://github.com/moq-wg/warp-streaming-format/issues/60#issuecomment-3359672623 was that a fragment response has a Mime-type associated with it

We can register a media type for MOQT, if that is necessary.

> that the fragment construct of pointing at a subset of a resource was being abused to code in additional information not related to the resource

The information is obviously related to the resource.  I don't understand this objection.

> flexible query arguments were considered a cleaner and more extensible fit

What if the application already uses some of those query parameters?  RFC 8820, Section 2.4 discourages/proscribes this kind of approach for this reason in some scenarios.

> For MSF, we need to give the client both a namespace and a name (as they are applied separately in all MOQT messages). So we then need to create formatting rules for the fragment to separate namespace from name.

I think Cullen has already attempted to do that in moq-wg/moq-transport#1355.

----------------------

Actually, now that I think more about this, is there anything MSF-specific in this proposal?  As far as I can tell, this basically explains how to access a MOQT track, so I do wonder if MOQT would be a more appropriate place for this.

michalhosna:
> I don't understand the circumstances under which this would be useful. Usually I would expect a MOQT resource to be available either via WT or raw QUIC, but not both.

I would expect that quite a lot of content may be available over both. Servers essentially have to support WebTransport for browsers. Supporting raw QUIC on the server is then straightforward, as one already has a QUIC stack anyway.
For clients, I would expect that non-browser clients could often implement QUIC only so as not to have an unnecessarily heavy HTTP stack. This is, of course, dependent on what we decide here.

> A URL "provide a means of locating the resource by describing its primary access mechanism" (RFC 3986), so not knowing the protocol sounds like a glaring omission.

`https://` currently means H1/H2/H3 with vastly different transports underlying it. I think it works fine in practice, and we do not have duplicate URLs representing the same resource. I do understand there is a fallback and discovery upgrade mechanism. 

That discovery works even better here. Both clients and servers can do ALPN, signaling their preferred and supported protocols. No race or additional RTTs required.

I do not like having two schemes, as that would create an ecosystem split. Which is, in my opinion, unfavorable, and in this case unnecessary given the high similarity of raw QUIC and WT. If the stated goal is to have compatibility with existing CMS systems, having a single URL representing a single piece of content is favorable. 

Also, what about WebTransport over H2? That one will have entirely different transport. Then having one scheme for raw QUIC and another for WT that would mean QUIC or TCP under the hood makes no sense to me. And having a different scheme from WT/H3 and WT/H2 would totally break a fallback use case. 

If we introduce more than one scheme/URL format, I would expect we'll have one transport (WT) in practice, and nobody will support the other ones. The proposed solution of a single URL allowing multiple transports seems not that different in that case. We could mandate relay/server support for WT in here. 

> I think Cullen has already attempted to do that in moq-wg/moq-transport#1355.

+1 from me on unifying format here with what comes out of 1355 (if it ends up URL compatible, which is one of the current goals there).

---- 

> Actually, now that I think more about this, is there anything MSF-specific in this proposal? As far as I can tell, this basically explains how to access a MOQT track, so I do wonder if MOQT would be a more appropriate place for this.

I think the problem is in the "content type". As we do not have, for now at least, a way to identify what kind of data are in the track, and for generic MOQT you cannot assume anything. This could be solved by multitude of ways, like having a content type extension/"header" in MOQT or having a custom scheme here that always represents a warp/msf catalog or track, but I do not think there is now a good way in MOQT itself.

vasilvv:
> https:// currently means H1/H2/H3 with vastly different transports underlying it. I think it works fine in practice, and we do not have duplicate URLs representing the same resource. I do understand there is a fallback and discovery upgrade mechanism.

There is an incredible amount of complexity involved in making a resource available over both H2 and H3.  Given that the advantage of making MOQ available over both raw QUIC and WebTransport is very limited, I cannot see it being justified here.

> That discovery works even better here. Both clients and servers can do ALPN, signaling their preferred and supported protocols. No race or additional RTTs required.

There is no guarantee that if a server offers both `moqt` and `h3` as ALPN, that the MOQT server in question would also be available over HTTP/3.  This also seems to ignore the TCP fallback for WebTransport.

> I do not like having two schemes, as that would create an ecosystem split.

The ecosystem split has already been created when we decided to support both raw QUIC and HTTP/3.

> If the stated goal is to have compatibility with existing CMS systems, having a single URL representing a single piece of content is favorable.

I agree.  In practice, I expect most deployments to pick one.

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+  the server in initializing and configuring the connection.
+* The query is optional. If present, it holds key-value data. Certain keys are reserved
+  by this specificaiton to hold special data. Multiple key/value pairs MAY be included,
+  separated by an ampersand. Query options MUST be stripped from the path when the path
+  is sent to the server at conneciton establishment. Query arguments are intended only
+  for the player and MUST NOT be transmitted to the server.
+
+An example URL is shown below:
+
+moqt://example.com/relay-app/relayID?c=customerID/broadcastID/catalog&c4m=12345
```
gwendalsimon, line 1195: `c` must be replaced by `ns`

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+* URL pointing at a subclip of a catalog
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID&location-range=34-0-64-16
+
+* URL pointing at a catalog and supplying a token
+  moqt://example.com/relay-app/relayID?ns=customerID/broadcastID
+  &c4m=gqhkYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5u
+  ZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM
+
+### Connection using WebTransport
+Assuming a WARP URL of
```
gwendalsimon, line 1255: MSF

draft-ietf-moq-warp.md
```diff
@@ -1171,6 +1172,117 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+1. The player generates the WebTransport URL by substituting https for moqt and stripping off
+   all query args. The token bytes are extracted and stored separately.
+2. Player establishes a Webtransport connection to 'https://example.com/relay-app/relayID' and
+   once connected, it executes CLIENT_SETUP supplying the token bytes via an AUTHORIZATION TOKEN
+   parameter and registering a token alias.
+3. Player subscribes to NAMESPACE customerID | broadcastID (where | represents the tuple
+   boundary) and to NAME catalog, along with the token alias.
+
+### Connection using raw QUIC
+Assuming a WARP URL of
```
gwendalsimon, line 1270: MSF

wilaw:
At the Boulder interim, the workgroup decided in favor of the fragment-based URl scheme, so closing this PR.

PR #87, opened 2026-01-16
Document fragment-based MSF URL structure

### LINKED ISSUE #60: URL syntax identifying both connection and track

### LINKED ISSUE #117: Fragment parameterization
IRT a fragment URL utilizing this style displays HTTP style querystring key/value pairs following the `?` and separation of `&`; then the fragment delimiter is `#` and its parameters which are also key/value pairs are additionally separated by `&`:

`moqt://example.com/moq?token=cafebeef0&abc=xyz#ns=broadcast/las&name=news`

wilaw:
Adds new section on MSF URL construction that is #fragment-based. 

Example MSF URLs:

URl with a required Webtransport connection pointing at a catalog track: 
`moqt+wt://example.com/server/config?a=1&b=2#customer-livestream-123--catalog`

Session: https://example.com/server/config?a=1&b=2
Namespace: ('customer', 'livestream', '123')
Track Name: 'catalog'

URL with a required raw QUIC connection pointing at a catalog: 
`moqt+q://example.com/relay-app/relayID#customerID-broadcastID--catalog`

URL pointing at a non-catalog track (either Webtransport or native QUIC may be used): 
`moqt://example.com/relay-app/relayID#customerID-broadcastID--video`

URL pointing at a subclip: 
`moqt://example.com/relay-app/relayID#customerID-broadcastID--catalog&location-range=34-64`

URL pointing at a catalog and supplying a token for the client: 
`moqt://example.com/relay-app/relayID#customerID-broadcastID--catalog&c4m=gqh kYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZlbWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmV qMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM`

URL pointing at a catalog and supplying a token for the client along with a separate token for the server: 
`moqt://example.com/relay-app/relayID?token=HTRCII74GHFT@JHBCVSW56HKKneH2Dbyq6NHB I2#customerID-broadcastID--catalog&c4m=gqhkYWxnIGVzaGFyqGR0eXBNhdZ9hdWQAY3VybGZl bWlzcwZleWV2aW5uZWlhdGVwQWNyZW5lYnJmcmVqMTIzNDU2NzgwMHZpc3VlZF9hdD0xNzMwNDM`

Fixes: #60 
Fixes: #117

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1246,7 +1246,7 @@ fragment, MSF uses a specific encoding convention:
 * Delimiter: The Track Name is appended to the end, separated from the namespace by
   a double hyphen (--).
 * Character Escaping:
-    * Unreserved characters [```a-z, A-Z, 0-9, _```] are represented literally.
+    * Unreserved characters (a-z, A-Z, 0-9, _) are represented literally.
     * All other byte values (including hyphens and periods used as data) MUST be
     * percent-encoded using a period (.) followed by two lowercase hexadecimal digits
       (e.g., a literal hyphen in a name becomes .2d).
@@ -1257,8 +1257,7 @@ Note: This encoding ensures that the structural delimiters (- and --) remain una
 The following ABNF defines the MSF URL structure, importing core rules from {{RFC5234}} and
 {{RFC3986}}.
 
-{:.alphabet}
-```abnf
+~~~ abnf
 msf-url         = msf-scheme "://" authority path-abempty [ "?" query ] [ "#" msf-fragment ]
 
 msf-scheme      = "moqt" / "https"
@@ -1270,7 +1269,8 @@ msf-track-name  = msf-encoded-string
 msf-encoded-string = *( unreserved / "." 2HEXDIG )
 unreserved      = ALPHA / DIGIT / "_"
 
-msf-query-arg   = wallclock-arg / mediatime-arg / location-arg / token-arg / other-arg
+msf-query-arg   = wallclock-arg / mediatime-arg /
+                  location-arg / token-arg / other-arg
 wallclock-arg   = "wallclock-range=" 1*DIGIT [ "-" 1*DIGIT ]
 mediatime-arg   = "mediatime-range=" 1*DIGIT [ "-" 1*DIGIT ]
 location-arg    = "location-range=" start-loc [ "-" end-loc ]
@@ -1279,7 +1279,7 @@ end-loc         = group-id "-" object-id
 group-id        = 1*DIGIT
 object-id       = 1*DIGIT
 token-arg       = "c4m=" 1*( ALPHA / DIGIT / "-" / "_" ) [ "=" / "==" ]
-```
+~~~
 
 ### Example MSF URLs
 
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -48,12 +48,15 @@ author:
 normative:
   MoQTransport: I-D.draft-ietf-moq-transport-11
   LOC: I-D.draft-mzanaty-moq-loc-05
+  C4M: I-D.draft-ietf-moq-c4m
   BASE64: RFC4648
   JSON: RFC8259
   LANG: RFC5646
   MIME: RFC6838
   RFC9000: RFC9000
   RFC4180: RFC4180
+  RFC5234: RFC5234
+  RFC3986: RFC3986
   GZIP: RFC1952
   WEBCODECS-CODEC-REGISTRY:
     title: "WebCodecs Codec Registry"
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+
+An MSF URL identifies a MOQT session and an optional sub-resource within that session.
+It follows the generic URI syntax:
+
+scheme : // authority / path ? query # fragment
+
+Component Definitions:
+* Scheme: Defines the underlying transport.
+    * moqt: Indicates a raw QUIC connection.
+    * https: Indicates a WebTransport connection.
```
michalhosna: For reasons [mentioned before](https://github.com/moq-wg/msf/issues/60#issuecomment-3335024567). I would still suggest a different scheme than HTTPS for the WT implementation.

(Suggestion, non-blocking, free-to-ignore)

draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+  as milliseconds since Unix Epoch and separated by a "-" dash. The dash and end
+  time MAY be omitted to indicate an open range.
+* mediatime-range - a range defined by start and end media times, each expressed
+  as milliseconds and separated by a "-" dash. The dash and end time MAY be omitted
+  to indicate an open range.
+* location-range - a range defined by start and end media MOQT Location tuples and
+  expressed as Start Group ID, Start Object ID, End Group ID, End Object ID, each
+  separated by a "-" dash. The End Group ID and End Object ID MAY be omitted to
+  indicate an open range in which the first two values are interpreted as
+  StartGroupID-StartObjectID.
```
michalhosna:
```suggestion
* location-range - a range defined by start and end media MOQT Location
  separated by a "-" dash. MOQT Location is expressed as Group ID and Object ID
  separated by a "." dot. End Location may be omitted to indicate an open range.
  End Object ID may be ommited, indicating the whole end group is included in
  the range. The "." dot and "-" dash separators MUST be omitted when the second value is ommited. 
```

I think separating 4 values of different types with the same character with the possibility of omitting some is confusing. I would fear that `StartGroupID-StartObjectID` may be easily confused for a case of `StartGroupID-EndGroupID`. Which I think also makes sense to support. Or at least support not sending the end object ID.

draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+* c4m - a base64 encoded token, as defined by {{C4M}}.
+
+Example query args:
+
+* wallclock-range=1761759637565-1761759836189
+* wallclock-range=1761751753894
+* mediatime-range=0-13421
+* mediatime-range=982
+* location-range=34-0-2145-16
+* location-range=16-24
```
michalhosna:
```suggestion
* location-range=34.0-2145.16
* location-range=16.24
```
If https://github.com/moq-wg/msf/pull/87/changes#r2700596913 is accepted

draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+  a single hyphen (-).
+* Delimiter: The Track Name is appended to the end, separated from the namespace by
+  a double hyphen (--).
+* Character Escaping:
+    * Unreserved characters (a-z, A-Z, 0-9, _) are represented literally.
+    * All other byte values (including hyphens and periods used as data) MUST be
+    * percent-encoded using a period (.) followed by two lowercase hexadecimal digits
+      (e.g., a literal hyphen in a name becomes .2d).
+
+Note: This encoding ensures that the structural delimiters (- and --) remain unambiguous.
```
michalhosna, line 1281: Is there a reason to not at least note that this is following [MOQT formatting recommendations]( 
https://www.ietf.org/archive/id/draft-ietf-moq-transport-16.html#name-representing-namespace-and-)?

draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+msf-encoded-string = *( unreserved / "." 2HEXDIG )
+unreserved      = ALPHA / DIGIT / "_"
+
+msf-query-arg   = wallclock-arg / mediatime-arg /
+                  location-arg / token-arg / other-arg
+wallclock-arg   = "wallclock-range=" 1*DIGIT [ "-" 1*DIGIT ]
+mediatime-arg   = "mediatime-range=" 1*DIGIT [ "-" 1*DIGIT ]
+location-arg    = "location-range=" start-loc [ "-" end-loc ]
+start-loc       = group-id "-" object-id
+end-loc         = group-id "-" object-id
```
michalhosna:
```suggestion
start-loc       = group-id "." object-id
end-loc         = group-id [ "." object-id ] 
```

Maybe also making Start object id optional with default value of 0 makes sense?

michalhosna (changes requested):
Thanks, Will! This addresses a lot of my nitpicks from https://github.com/moq-wg/msf/pull/72, sorry for them.
I do expect some MSF URL definitions may need additional work once the text on the URL handling exists.

draft-ietf-moq-msf.md
```diff
@@ -1155,6 +1174,136 @@ This example shows drone GPS coordinates synched with the start of each Group.
 
 # Workflow
 
+## URL construction and interpretation
+
+An MSF URL identifies a MOQT session and an optional sub-resource within that session.
+It follows the generic URI syntax:
+
+scheme : // authority / path ? query # fragment
```
suhasHere: For my own understanding, any reason why we moved away from query parameter. Fragments have its own shortcomings.

mondain:
Idea for fragment parameters https://github.com/moq-wg/msf/issues/117

draft-ietf-moq-msf.md
wilaw: We discussed this issue at Boulder interim. Consensus was that fragment approach was more aligned with the behavior of a player in which the namespace info is not communicated to the relay during the initial connection.

draft-ietf-moq-msf.md
wilaw: We will inherit the new scheme which MOQT has just implemented via https://github.com/moq-wg/moq-transport/pull/1486

draft-ietf-moq-msf.md
wilaw: We will inherit the new scheme defined by https://github.com/moq-wg/moq-transport/pull/1486

draft-ietf-moq-msf.md
```diff
@@ -1156,6 +1174,142 @@ This example shows drone GPS coordinates synched with the start of each Group.
...
+with optional key-value attributes.
+
+"moqt" [ "+q" / "+wt" ] "://" authority path-abempty [ "?" query ] [ "#" msf-fragment ]
+
+The MOQT specification carries the normative definition of these components,
+along with processing instructions. They are repeated here for clarity:
+
+* Scheme: This case-insensitive scheme defines the underlying transport.
+    * moqt: the client may use either a WebTransport or native QUIC connection.
+    * moqt-q: the client MUST use a native QUIC connection.
```
suhasHere:
```suggestion
    * moqt+q: the client MUST use a native QUIC connection.
```

PR #88, opened 2026-01-18
Fix typos and JSON syntax errors

suhasHere:


wilaw (approved):
These look good. Thanks for catching.

PR #114, opened 2026-02-08
Fix typos and clarifications

ac-freeman:


wilaw (approved):
Thanks for making these fixes.

PR #118, opened 2026-02-11
add details of authorization flows

### LINKED ISSUE #119: Add authz details

suhasHere:
Fixes #119

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -725,6 +725,28 @@ the track name and other parameters defined in ({{SecureObjects, Section 5}}),
 is used to derive the actual encryption keys. Publishers and subscribers MUST
 use matching trackBaseKey values for successful decryption.
 
+### Authorization Schemes {#authschemes}
+Location: R    Required: Optional    JSON Type: Array
+
+An array of strings indicating the authorization mechanisms supported by the
+publisher for accessing tracks in this catalog. Subscribers MUST authenticate
+using one of the advertised schemes before accessing protected tracks.
+
+Table 7: Registered Authorization Schemes
+
+| Scheme           | Value          | Reference                              |
+|:================|:===============|:======================================|
+| Privacy Pass     | privacy-pass   | {{PrivacyPassAuth}}                    |
+| CAT              | cat            | {{C4M}}                                |
+
+Publishers MAY support multiple authorization schemes simultaneously. When
+multiple schemes are listed, they are presented in order of publisher preference.
+A subscriber SHOULD use the first scheme it supports from the list.
+
+Custom authorization schemes MAY be used. Custom scheme names MUST use a
+unique naming convention, such as Reverse Domain Name Notation
+(e.g., "com.example.custom-auth"), to avoid naming collisions.
+
 ## Delta updates {#deltaupdates}
 A catalog update might contain incremental changes. This is a useful property if
 many tracks may be initially declared but then there are small changes to a
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -752,9 +752,27 @@ scheme to use and then obtain tokens through out-of-band mechanisms.
 The specific token format, acquisition process, and presentation method
 are defined by the authorization scheme specification.
 
-TODO: Consider adding a mechanism to indicate which token should be applied
-during connection establishment vs. per-track operations. Revisit once MOQT
-variable parameters support is finalized.
+### Token Delivery via URI {#tokendelivery}
+
+Tokens can be delivered to subscribers through the catalog request URI using
+variable substitution. Fragment parameters (following the `#` character) are
+processed client-side and can be substituted into catalog fields using the
+variable substitution mechanism defined in {{variablesubstitution}}.
+
+For example, given a URI:
+
+    moqt://relay.example.com/live#token=XYZ789
+
+A catalog with:
+
+~~~json
+"authInfo": {
+  "cat": "%token%"
+}
+~~~
+
+Would be resolved by the subscriber to include `"cat": "XYZ789"`, which is
+then presented in control messages as specified by the authorization scheme.
 
 ## Delta updates {#deltaupdates}
 A catalog update might contain incremental changes. This is a useful property if
@@ -1509,7 +1527,11 @@ Would be resolved by the subscriber as:
 ### Time-aligned Audio/Video Tracks with Authorization
 
 This example shows a catalog for a media producer requiring authorization
-for premium content.
+for premium content. The premium 4K track uses CAT authorization with a
+token resolved via variable substitution from `%cat-token%`. The standard
+720p track uses Privacy Pass with a token resolved from `%pp-token%`.
+Token presentation timing is determined by the authorization scheme
+specification.
 
 ~~~json
 {
@@ -1530,7 +1552,7 @@ for premium content.
       "framerate": 60,
       "bitrate": 15000000,
       "authInfo": {
-        "cat": {}
+        "cat": "%cat-token%"
       }
     },
     {
@@ -1547,7 +1569,7 @@ for premium content.
       "framerate": 30,
       "bitrate": 2500000,
       "authInfo": {
-        "privacy-pass": {}
+        "privacy-pass": "%pp-token%"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -516,6 +519,54 @@ Location: T    Required: Optional   JSON Type: Number
 The duration of the track expressed in integer milliseconds. This field MUST NOT
 be included if the isLive {{islive}} field value is true.
 
+### Authorization Schemes {#authschemes}
+Location: R    Required: Optional    JSON Type: Array
+
+An array of strings indicating the authorization mechanisms supported by the
+publisher for accessing tracks in this catalog. Subscribers MUST authenticate
+using one of the advertised schemes before accessing protected tracks.
```
wilaw: Why is this necessary? The player can inspect the tracks and determine the same information. I think we can do without this redundant information in the root of the catalog.

draft-ietf-moq-msf.md
```diff
@@ -516,6 +519,54 @@ Location: T    Required: Optional   JSON Type: Number
...
+| issuerUrl       | String  | URL of the token issuer                        |
+| scope           | String  | Required scope pattern for token validation    |
+
+For CAT authentication (scheme "cat"), the object MAY contain:
+
+| Field           | Type    | Description                                    |
+|:===============|:========|:==============================================|
+| actions         | Array   | Required MOQT action codes (0-8)               |
+| revalInterval   | Number  | Token revalidation interval in seconds         |
+| dpopRequired    | Boolean | Whether DPoP proof binding is required         |
```
wilaw: The player is not going to construct its own CAT token, therefore I don't think it needs this information. The internals of the CAT token are opaque to the player. It only needs to be told which token to apply to which track, and also whether it needs to apply the CAT token to the relay when establishing the initial connection. The current proposal does not provide this information.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
   publish an independent catalog update which signals isComplete {{iscomplete}} as
   TRUE and which contains an empty Tracks {{tracks}} field.
 
+## Authorization {#authorization}
+
+MSF supports token-based authorization through pluggable authentication schemes.
+Both original publishers and end subscribers can independently acquire authorization
+tokens and present them to relays. Relays use these tokens to authorize whether
+an end subscriber is permitted to receive content and whether an original publisher
+is permitted to send content.
```
wilaw: The CAT token can also authorize the initial establishment of the MOQT or WebTransport connection.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+
+### Discovering Authorization Requirements
+
+End subscribers discover authorization requirements by parsing the catalog:
+
+1. Check if `authSchemes` is present at the catalog root level
```
wilaw: We can omit this step

gwendalsimon (approved):
Awesome!

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+2. For each track of interest, examine `authInfo` for scheme-specific details
+3. Select a supported scheme from `authSchemes` and obtain appropriate credentials
+
+Original publishers and end subscribers may obtain authorization tokens from
+multiple sources, including the catalog's `authInfo` field or as a parameter
+in the connection URI. These tokens can be presented either in the connection
```
wilaw: We need to define a standard way for transmitting tokens to the player. It will most likely be via query args in the URL, combined with the variable syntax proposed in the Boulder interim.

`https://example.com/moqt#my-protected-stream--catalog&token1=12345&token2=6789`

In catalog for a track entry:

```
"bitrate": 15000000,
"authInfo": {
     "cat": %token1%
 },
```

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+setup message (CLIENT_SETUP) or in control messages (SUBSCRIBE, FETCH, PUBLISH,
+PUBLISH_NAMESPACE) as described in {{presentingauthorization}}.
+
+### Token Acquisition
+
+Token acquisition is out of scope for this specification. Both original publishers
```
wilaw, line 2302: As indicated above, we should define a standard way of presenting tokens using base64 encoded tokens in the query portion of the URL.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+
+### Presenting Authorization {#presentingauthorization}
+
+Once tokens are obtained, both original publishers and end subscribers present
+them to relays according to the scheme specification. Tokens MAY be presented
+at two points during the connection lifecycle:
+
+* **Connection establishment**: Tokens may be included in the SETUP message
+  (CLIENT_SETUP or SERVER_SETUP) using the AUTHORIZATION TOKEN setup parameter.
+  This authorizes the connection itself or pre-authorizes subsequent operations.
```
wilaw: How do we tell the player to send a token during connection? We can say that if a player is given a token that it always presents it during connection establishment, as the relay will ignore tokens it is not configured to process. Or we can invent a root catalog entry to indicate that a token should be provided during connectionTime:

"connectionToken": "%token1%"

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+
+* **Per-operation**: Tokens may be included in individual control messages
+  using the AUTHORIZATION TOKEN message parameter. For end subscribers, this
+  applies to SUBSCRIBE, SUBSCRIBE_NAMESPACE, FETCH, and REQUEST_UPDATE messages. For original
+  publishers, this applies to PUBLISH and PUBLISH_NAMESPACE messages. When a track's catalog entry includes an
```
wilaw: Is this a normative 'may' ?

I think we can simplify things by saying that any time a token is matched with a track, that it should then be provided for ALL control messages which accept token parameters and are associated with that track.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+
+* **Original publishers** include an AUTHORIZATION TOKEN parameter in CLIENT_SETUP
+  when connecting to a relay. This token MAY pre-authorize the publisher
+  for subsequent PUBLISH and PUBLISH_NAMESPACE operations on namespaces
+  covered by the token's scope.
+
+The SETUP token establishes a default authorization context for the session.
+If a relay requires authorization and no token is provided in SETUP, the
+endpoint MUST provide valid authorization in each individual control message
+that requires it.
```
wilaw: Even if a token is sent during SETUP, it (or its alias) MUST still be resent in all subsequent messages which require token auth.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+
+### Token Override in Control Messages
+
+When an endpoint includes an AUTHORIZATION TOKEN parameter in an individual
+control message, that token overrides any token previously established in the
+SETUP message for that specific operation. This override mechanism enables
+several use cases:
```
wilaw: I don't think the MOQT currently works like this. Tokens supplied during setup are not automatically applied to subsequent messages in that session.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+The specific token format and placement depends on the authorization scheme:
+
+* **Privacy Pass**: Tokens are included in MOQT control messages
+  (SUBSCRIBE, FETCH, PUBLISH, PUBLISH_NAMESPACE) via the AUTHORIZATION TOKEN
+  parameter with auth_scheme 0x01. Connection-level tokens may be included in
+  the SETUP message. See {{PrivacyPassAuth}} for details.
+
+* **CAT**: Tokens are supplied via MOQT's AUTHORIZATION TOKEN parameter.
+  When DPoP is required, each operation includes a fresh DPoP proof JWT
+  bound to the endpoint's public key. See {{CATAuth}} for details.
```
wilaw: This whole section is just informative and is repeating information already supplied in MOQT draft, C4M draft and Privacy pass draft. We should remove it from the MSF draft.

draft-ietf-moq-msf.md
```diff
@@ -1173,6 +1294,148 @@ is complete by taking the following steps:
...
+6. Reject requests that fail any validation step
+
+For original publishers sending content via PUBLISH or PUBLISH_NAMESPACE, relays MUST:
+
+1. Verify the publisher has presented valid credentials
+2. Validate the token cryptographically
+3. Check that the token scope authorizes publishing to the namespace or track
+4. Reject announcements or publications that fail any validation step
+
+Relays SHOULD return an appropriate error if authorization fails.
```
wilaw: Again, this is repeating info that is supplied in the MOQT, C4M and PP specs. We should nto include it here.

What we should include are some instructions for a MSF player on what to do when a token is rejected by the relay and an authentication error is returned.

wilaw (commented):
Thank you for this PR. It is a much needed addition to the draft.

I think there are several sections where we can simplify both the catalog entries and the workflow. Also can remove a bunch of informative prose from this PR. Plus the workflow needs to be refined and we need to tell the players how to extract tokens from query args, which is likely to be the most common way of supplying tokens to players.

draft-ietf-moq-msf.md
suhasHere: Agree. I have another PR addressing this and happy to update this section once that PR lands.

draft-ietf-moq-msf.md
```diff
@@ -1340,6 +1373,67 @@ Would be resolved by the subscriber as:
...
+      "targetLatency": 2000,
+      "role": "video",
+      "renderGroup": 1,
+      "codec": "av01.0.12M.10.0.110.09",
+      "width": 3840,
+      "height": 2160,
+      "framerate": 60,
+      "bitrate": 15000000,
+      "authInfo": {
+        "cat": {}
```
wilaw: Wouldn't this be a string containing a base64 serialized token, instead of a JSON Object?

Something like   "cat": "%cat-token%"

draft-ietf-moq-msf.md
```diff
@@ -1340,6 +1373,67 @@ Would be resolved by the subscriber as:
...
+      "targetLatency": 2000,
+      "role": "video",
+      "renderGroup": 2,
+      "codec": "av01.0.05M.10.0.110.09",
+      "width": 1280,
+      "height": 720,
+      "framerate": 30,
+      "bitrate": 2500000,
+      "authInfo": {
+        "privacy-pass": {}
```
wilaw: same here. Wouldn't this be a string containing a base64 serialized token, instead of a JSON Object?

Something like   "privacy-pass": "%pp-token%"

draft-ietf-moq-msf.md
```diff
@@ -2162,6 +2276,88 @@ is complete by taking the following steps:
...
+The token acquisition endpoint and flow depend on the authorization scheme
+and deployment configuration. Original publishers and end subscribers MAY
+use the same or different authorization services depending on the deployment.
+
+Tokens MAY also be supplied as parameters in the connection URI. When a
+connection URI contains token parameters, the subscriber extracts the token
+value and includes it in the appropriate MOQT message. For example:
+
+~~~
+https://relay.example.com/moqt?token=eyJhbGciOiJFZERTQSJ9...
```
```suggestion
moqt://relay.example.com/moqt#namespace--name&token=eyJhbGciOiJFZERTQSJ9...
```

draft-ietf-moq-msf.md
```diff
@@ -725,6 +725,55 @@ the track name and other parameters defined in ({{SecureObjects, Section 5}}),
...
+### Token Delivery via URI {#tokendelivery}
+
+Tokens can be delivered to subscribers through the catalog request URI using
+variable substitution. Fragment parameters (following the `#` character) are
+processed client-side and can be substituted into catalog fields using the
+variable substitution mechanism defined in {{variablesubstitution}}.
+
+For example, given a URI:
+
+    moqt://relay.example.com/live#token=XYZ789
```
```suggestion
    moqt://relay.example.com/live#namespace--name&token=XYZ789
```

draft-ietf-moq-msf.md
```diff
@@ -2162,6 +2276,88 @@ is complete by taking the following steps:
   publish an independent catalog update which signals isComplete {{iscomplete}} as
   TRUE and which contains an empty Tracks {{tracks}} field.
 
+## Authorization {#authorization}
+
+MSF supports token-based authorization through pluggable authentication schemes.
+Both original publishers and end subscribers can independently acquire authorization
+tokens and present them to relays. Relays use these tokens to authorize whether
+an end subscriber is permitted to receive content and whether an original publisher
+is permitted to send content.
```
```suggestion
tokens and present them to relays. Relays use these tokens to authorize whether
an end subscriber is permitted to setup the connection, as well as to receive content
and/or to send content.
```

wilaw (approved):

PR #122, opened 2026-02-19
initial text on zapping

### LINKED ISSUE #109: Alternative tracks with different periods of random access points
Think about what changes are needed to MSF (or CMSF) for support of alternative tracks with different periods of random access points for enabling track switching delay optimization.

### LINKED ISSUE #110: zapping and quick switching
What should we put in catalog to enable tracks with I-frames

suhasHere:
Fixes #110 
Fixes #109


draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
 
 Each subsequent Group ID MUST increase by 1.
 
 If a publisher is able to maintain state across a republish, it MUST signal the gap
 in Group IDs using the MOQT Prior Group ID Gap Extension header.
 
+## Zapping and Track Switching {#zapping}
+
+Zapping refers to the act of quickly switching between media streams, such as
+changing channels in a live broadcast. When a viewer switches to a new track,
```
```suggestion
Zapping refers to the act of quickly joining a track or switching between tracks, such as changing channels in a live broadcast. When a viewer switches to a new track,
```

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+enables subscribers to calculate these valid switch points.
+
+### Calculating Valid Switch Points
+
+When switching from track A (with spacing N) to track B (with spacing M), a
+subscriber can switch at group ID_A in track A if ((ID_A - 1) * N / M) + 1 yields
+an integer result. The resulting integer is the target group ID in track B.
+
+All tracks within the same alternate group that specify a trackSpacing value
+MUST be time-aligned at object boundaries, not only at group boundaries.
```
wilaw: Also, aif one track within an altGroup specifies a trackSpacing value then all tracks in theat altGroup MUST specify a trackSpacing value.

draft-ietf-moq-msf.md
```diff
@@ -517,6 +517,16 @@ Location: T    Required: Optional   JSON Type: Number
 The duration of the track expressed in integer milliseconds. This field MUST NOT
 be included if the isLive {{islive}} field value is true.
 
+### Track spacing {#trackspacing}
```
wilaw: Can we call this 'keyframeSpacing' ? That is a little more precise than trackSpacing.

draft-ietf-moq-msf.md
wilaw: In the https://moq-wg.github.io/msf/wilaw-patch-4/draft-ietf-moq-msf.html#name-media-packaging section of the doc, add a section which instructs publishers how to properly package content with dissimilar keyframe spacing. It is also important that Groups are placed at every keyframe boundaries. If you put 2 keyframes inside one group, then intermediate join points will be hidden.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+
+Consider two video tracks in an alternate group:
+
+* Track A (HD): trackSpacing = 1
+* Track B (SD): trackSpacing = 2 ( twice the number of objects )
+
+The following diagram shows the group structure and valid switch points:
+
+~~~ascii-figure
+Track A (spacing=1):
```
wilaw: My formula was developed here: https://docs.google.com/presentation/d/1COTZBM3oiYbVC_XR8R25aHYRPlz635RdRUAGNj150M8/edit?slide=id.g3bd55a03409_0_227#slide=id.g3bd55a03409_0_227

It defines the track spacing as "the number of objects per keyframe". 

IN this example for track A, there are 2 objects per keyframe, so the spacing should be 2.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+  Group:    1     2     3     4     5     6     7     8     9
+            |     |     |     |     |     |     |     |     |
+           [K]   [P]   [K]   [P]   [K]   [P]   [K]   [P]   [K]
+            ^           ^           ^           ^           ^
+            |           |           |           |           |
+         keyframe    keyframe   keyframe   keyframe   keyframe
+
+
+Track B (spacing=2):
```
wilaw: Spacing should be 4 for track B as there a 4 objects per keyframe.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+Consider two video tracks in an alternate group:
+
+* Track A (HD): trackSpacing = 1
+* Track B (SD): trackSpacing = 2 ( twice the number of objects )
+
+The following diagram shows the group structure and valid switch points:
+
+~~~ascii-figure
+Track A (spacing=1):
+  Group:    1     2     3     4     5     6     7     8     9
```
wilaw: These groups are wrong. The group boundaries should match the **keyframe** boundaries.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+            |     |     |     |     |     |     |     |     |
+           [K]   [P]   [P]   [P]   [K]   [P]   [P]   [P]   [K]
+            ^                       ^                       ^
+            |                       |                       |
+         keyframe               keyframe               keyframe
+
+
+Valid switch points (A -> B):
+  - Group 1 in A -> Group 1 in B  ((1-1)*2/4)+1 = 1    (valid)
+  - Group 3 in A -> Group 2 in B  ((3-1)*2/4)+1 = 2    (invalid, B[2] is P-frame)
```
wilaw: This is wrong. Object 3 in track A would be Gorup ID 2. It's the gorup ID that must be put in the forumla, not the object count.

So switching from group 2 in tracks A yields ((2-1)*2/4)+1 = 1.5, therefore invalid since it is not an integer.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+  - Group 5 in A -> Group 3 in B  ((5-1)*2/4)+1 = 3    (invalid, B[3] is P-frame)
```
wilaw: Switrching from Group 3 in track A yields ((3-1(*2/4)+1 = 2, so it is valid and can switch to Gorup 2 in track B.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+  Group:    1     2     3     4     5     6     7     8     9
+            |     |     |     |     |     |     |     |     |
+           [K]   [P]   [K]   [P]   [K]   [P]   [K]   [P]   [K]
+            ^           ^           ^           ^           ^
+            |           |           |           |           |
+         keyframe    keyframe   keyframe   keyframe   keyframe
+
+
+Track B (spacing=2):
+  Group:    1     2     3     4     5     6     7     8     9
```
wilaw: Groups must match the k**eyframe boundaries**.

draft-ietf-moq-msf.md
```diff
@@ -1005,15 +1015,81 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+            ^                       ^                       ^
+            |                       |                       |
+         keyframe               keyframe               keyframe
+
+
+Valid switch points (A -> B):
+  - Group 1 in A -> Group 1 in B  ((1-1)*2/4)+1 = 1    (valid)
+  - Group 3 in A -> Group 2 in B  ((3-1)*2/4)+1 = 2    (invalid, B[2] is P-frame)
+  - Group 5 in A -> Group 3 in B  ((5-1)*2/4)+1 = 3    (invalid, B[3] is P-frame)
+  - Group 9 in A -> Group 5 in B  ((9-1)*2/4)+1 = 5    (valid)
```
wilaw: Should be Group 5 in A yields ((5-1)*2/4)+1 = 3, so can switch to group 3 in B. 

You should also show examples for switching down - i.e from B to A.

wilaw (changes requested):
Thanks for adding this! However, some logic and definition errors here which need resolving. The group boundaries must be aligned with the keyframes for my switching formula to work. The examples as shown are not valid, as the algorithm states that any integer result is valid, yet even you invalid examples show integer results. I marked it uo to show what needs fixing.

suhasHere:
@wilaw I think i have addressed your feedback in the last commit. Please give another read and see if it makes sense. Thanks

gwendalsimon:
I find this PR over-explaining something that is quite simple in theory. Do we need so many examples for such a basic thing?

The issue #110 was raised during the Boulder meeting (i was only the notetaker). The idea was to mimic a mechanism such as HESP, where one track has much more keyframes than other tracks, to enable fast switching/zapping. In this regard, the parameter in the manifest helps notifying the Subscriber about which track to use for fast joining. This is good.

However, the text does not say how to use it. The idea is to first join such a "many-I track", and then to quickly resume in a track with a lower frequency of keyframes, i.e. a zapping procedure is SUBSCRIBE track-with-many-keyframes latestObject, and, once Subscriber receives SUBSCRIBE_OK with a group number (say 32), the Subscriber can SUBSCRIBE track-with-less-keyframes Start=34 and REQUEST_UPDATE track-with-many-keyframes end=33. Thus, the track with many keyframes is used only for starting fast but not for staying.

suhasHere:
@gwendalsimon  we can simplify text if needed, but on how do we use aspect, i wonder how much of that should be in this spec ?

may be an appendix section ?

wilaw:
Image to help with examples form IETF #126
Screenshot 2026-07-30 at 18 44 57

Issue #106, opened 2026-02-05
a list of issues from 2026/02/05 MSF meeting at Denver


gwendalsimon:
- renderGroup is an integer. Should it be an array for tiling?
- add a default track from an altGroup (for example to select one default audio language from a set of audio tracks)
- instead of a flat list of tracks, define first a list of renderGroups, which defines each renderGroup parameters (e.g. targetLatency) + a list of tracks, each track pointing  to one of the defined renderGroups. Parameters can appear either in the renderGroup or in the track (but no in both).
- altGroup could also be defined in a list of altGroup, and each track points to a defined altGroup
- parameters like width/height/bitrate may be indicated as being "hints" (or max values) -- an option could be to modify these parameters with a delta update
- media timeline "depends" on a renderGroup rather than depending on a media track
- media timeline must accommodate that tracks can appear and be removed
- invent a mechanism to have a rolling buffer, to implement cDVR window (in the media timeline)
- invent a notion of template for media timeline

wilaw:
@wilaw  - triage and create separate issues if necessary.

wilaw:
Created these separate issues

renderGroup is an integer. Should it be an array for tiling? https://github.com/moq-wg/msf/issues/201
add a default track from an altGroup https://github.com/moq-wg/msf/issues/202
media timeline "depends" on a renderGroup rather than depending on a media track https://github.com/moq-wg/msf/issues/203
Media timeline must accommodate that tracks can appear and be removed https://github.com/moq-wg/msf/issues/204
How do we implement DVR with MSF? https://github.com/moq-wg/msf/issues/205

> instead of a flat list of tracks, define first a list of renderGroups, which defines each renderGroup parameters (e.g. targetLatency) + a list of tracks, each track pointing to one of the defined renderGroups. Parameters can appear either in the renderGroup or in the track (but no in both).

This was discussed at IETF #126. Feedback was that flexibility for future structures outweighed parsing gains, so keep it flat for now. 

>  
 invent a notion of template for media timeline

This was added with PR #127 

> parameters like width/height/bitrate may be indicated as being "hints" (or max values) -- an option could be to modify these parameters with a delta update

This was added with PR #187

PR #123, opened 2026-02-19
Add support for variable substitution

suhasHere:
Addresses  part of #106

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -227,6 +227,7 @@ Table 1 provides an overview of all fields defined by this document.
 | Language                | lang                   | {{language}}              |
 | Parent name             | parentName             | {{parentname}}            |
 | Track duration          | trackDuration          | {{trackduration}}         |
+| Variables               | variables              | {{variables}}             |
 
 Table 2 defines the allowed locations for these fields within the document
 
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -542,12 +542,48 @@ The following rules are to be followed in constructing and processing delta upda
...
+each parameter name becomes available as a variable. The variable is replaced
+with the corresponding parameter value.
+
+### Variables {#variables}
+Location: R    Required: Optional    JSON Type: Array
+
+An array of strings declaring the variable names used within this catalog.
+Publishers SHOULD include this field to enable validation and to document
+the expected variables. Each string in the array is a variable name without
+the enclosing percent characters.
```
wilaw: We can omit this declaration of variable from the catalog. It is not providing any new information. A parser can easily find all the instance of %*% in the document and the original publishwer is explicitly declaring variables by placing them within % fields. The text also does not explain what do to if the variables declared in this block are not the variables it finds in track properties. Can we start with removing it for simplicity and then add it in later if implementation experience shows that we need it?

draft-ietf-moq-msf.md
```diff
@@ -542,12 +542,48 @@ The following rules are to be followed in constructing and processing delta upda
...
+
+Variables are resolved from the URI used to access the catalog. The URI
+components used for variable resolution are:
+
+* Query parameters - the name and value pairs following the `?` character
+* Fragment identifier - the portion following the `#` character
+
+When a subscriber requests a catalog using a URI containing query parameters,
+each parameter name becomes available as a variable. The variable is replaced
+with the corresponding parameter value.
```
```suggestion
* Fragment parameters - the key/value parameters encoded into the fragment portion of the URL.

When a subscriber requests a catalog using a URL containing fragment parameters, each key name becomes available as a variable. The variable is replaced with the corresponding parameter value.
```

Updating to be consistent with #87

draft-ietf-moq-msf.md
```diff
@@ -991,6 +1027,64 @@ live broadcast containing a video and an audio track.
 
 ~~~
 
+### Variable Substitution for personalized delivery
+
+This example shows a catalog using variable substitution to enable
+personalized advertising and reporting while maintaining cacheability.
+Given a catalog request URI of:
+
+    https://example.com/relay-app/relayID?token=1234&id=bob&event=xyz
```
```suggestion
    moqt://example.com/relay-app/relayID?a=1&b=2#namespace--name&token=1234&id=bob&event=xyz
```
Note that 'a' and 'b' are not available as variables, as they are intended for the server and not for the client. Only parameters declared in the fragment can be used as variables.

wilaw (changes requested):
Add a few comments. I think we don;t need to doubnle declare the variables.

Also, what about variable safety? Do we open ourselves to injection attacks? 

moqt://example.com/relay-app/relayID?#namespace--name&myvar=,malicious-json-payload="bad stuff"

perhaps we limit variable values to exclude ,;'"&? in addition to just %. We could allow only 0..9 a..zA..Z_-@ etc.

suhasHere:
@wilaw  please see if the changes are fine. thanks

draft-ietf-moq-msf.md
```diff
@@ -991,6 +1026,63 @@ live broadcast containing a video and an audio track.
 
 ~~~
 
+### Variable Substitution for personalized delivery
+
+This example shows a catalog using variable substitution to enable
+personalized advertising and reporting while maintaining cacheability.
+Given a catalog request URI of:
+
+    moqt://relay.example.com/live-sports/catalog#token=1234&id=bob&event=xyz
```
```suggestion
    moqt://relay.example.com/sports/catalog?a=1#token=1234&id=bob&event=xyz
```

wilaw (approved):
Thanks for the changes. Looks good.

PR #124, opened 2026-02-19
Clarify first object in event and media timeline track

### LINKED ISSUE #103: Groups and objects of the media timeline track and the event timeline track
Some specifications, including constraints, of groups and objects of the media timeline track and the event timeline track seem needed. For example, the 1st object in a group needs to be self-contained thus contain the complete information (up to that point), while other objects in a group can be deltas, each containing a set of consecutive lines.

suhasHere:
Fixes #103

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1054,9 +1054,12 @@ The mime-type of a media timeline track MUST be specified as "application/json".
 
 ## Media Timeline track updating
 The publisher MUST publish an independent media timeline in the first MOQT Object
-of each MOQT Group of a media timeline track. The publisher MAY publish incremental
-updates in the second and subsequent Objects within each Group. Incremental updates
-only contain media timeline records since the last media timeline Object.
+of each MOQT Group of a media timeline track. An independent media timeline object
+MUST contain all media timeline records accumulated up to that point, allowing a
+subscriber joining at any group boundary to receive the complete timeline history.
+The publisher MAY publish incremental updates in the second and subsequent Objects
+within each Group. Incremental updates contain only new media timeline records since
+the previous media timeline Object in that Group.
 
 # Event Timeline track {#eventtimelinetrack}
 The event timeline track provides a mechanism to associate ad-hoc event metadata with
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -1055,9 +1054,12 @@ The mime-type of a media timeline track MUST be specified as "application/json".
 
 ## Media Timeline track updating
 The publisher MUST publish an independent media timeline in the first MOQT Object
-of each MOQT Group of a media timeline track. The publisher MAY publish incremental
-updates in the second and subsequent Objects within each Group. Incremental updates
-only contain media timeline records since the last media timeline Object.
+of each MOQT Group of a media timeline track. An independent media timeline object
+MUST contain all media timeline records accumulated up to that point, allowing a
```
gwendalsimon: Should we force "all records up to that point"? It seems excessive, for very long stream (cricket games) and even worse for 24/7 channels. It would be good to have a kind of "horizon", and so we would let the publisher send only the XX latest records, corresponding to now - horizon. We could add an optional parameter specific to both timelines to express the horizon (either in ms or number of groups). If not, then we would assume infinite horizon.

suhasHere: good point @gwendalsimon 
Also i wonder, the publisher application is the final decision maker. I wonder how would the marker help the subscriber to act on it ?

draft-ietf-moq-msf.md
```diff
@@ -1055,9 +1054,12 @@ The mime-type of a media timeline track MUST be specified as "application/json".
...
+subscriber joining at any group boundary to receive the complete timeline history.
```
```suggestion
MUST contain all media timeline records accumulated and accessible up to that point, allowing a
subscriber joining at any group boundary to receive the accessible timeline history.
```

draft-ietf-moq-msf.md
```diff
@@ -1099,9 +1101,12 @@ An event timeline track MUST carry:
 
 ## Event Timeline track updating
 The publisher MUST publish an independent event timeline in the first MOQT Object
-of each MOQT Group of an event timeline track. The publisher MAY publish incremental
-updates in the second and subsequent Objects within each Group. Incremental updates
-only contain event timeline records since the last event timeline Object.
+of each MOQT Group of an event timeline track. An independent event timeline object
+MUST contain all event timeline records accumulated up to that point, allowing a
+subscriber joining at any group boundary to receive the complete event history.
```
```suggestion
MUST contain all event timeline records accumulated and accessible up to that point, allowing a
subscriber joining at any group boundary to receive the accessible event history.
```

wilaw (commented):
Made two suggested changes to accommodate Gwendal's desire to support very long events. We can refine this approach later to a more sophisticated DVR scheme.

PR #126, opened 2026-02-19
Clarify delta processing fields require deltaUpdate to be true

### LINKED ISSUE #92: The delta processing instruction fields MUST not be present when the deltaUpdate field is false.
In the end of the following sections:

- 5.1.3. Add tracks
- 5.1.4. Remove tracks
- 5.1.5. Clone tracks

add the following sentence (or state the same in Section 5.1.2 for all these fields):
> This field MUST NOT be present in a catalog when the deltaUpdate field is false.

suhasHere:
Fixes #92


wilaw (approved):
LGTM

PR #127, opened 2026-02-19
Define template for Media Timeline track

### LINKED ISSUE #105: Think about template based media timeline track design
Such that for media tracks that are of constant durations the corresponding media timeline tracks can be of very small size.

suhasHere:
Fixes #105

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1078,7 +1078,11 @@ can exist inside a catalog.
 
 ## Media Timeline track payload {#mediatimelinepayload}
 A media timeline track is a JSON {{JSON}} document. This document MAY be compressed
-using GZIP {{GZIP}}. The document contains an array of records. Each record consists of
+using GZIP {{GZIP}}. The document supports two formats: an explicit entry format
+and a template format. Publishers MAY combine both formats in a single document.
+
+### Explicit entry format {#explicitentryformat}
+The explicit format contains an array of records. Each record consists of
 an array of three required items, whose ordinal position defines their type:
 
 * The first item holds the media presentation timestamp, expressed as a JSON Number.
@@ -1087,12 +1091,12 @@ an array of three required items, whose ordinal position defines their type:
 * The second item holds the MOQT Location of the entry, defined as a tuple of the MOQT
   Group ID and MOQT Object ID, and expressed as a JSON Array of Numbers, where the
   first number is the Group ID and the second number is the Object ID.
-* The third time holds the wallclock time at which the media was encoded, defined as
+* The third item holds the wallclock time at which the media was encoded, defined as
   the number of milliseconds that have elapsed since January 1, 1970
   (midnight UTC/GMT) and expressed as a JSON Number. For VOD assets, or if the
   wallclock time is not known, the value SHOULD be 0.
 
-An example media timeline is shown below:
+An example media timeline using explicit entries is shown below:
 
 ~~~json
 [
@@ -1103,6 +1107,66 @@ An example media timeline is shown below:
   [8008, [4,0], 1759924166389]
 ]
 ~~~
+
+### Template format {#templateformat}
+For media tracks with constant duration GOPs, publishers MAY use a template format
+to reduce payload size. The template format uses a JSON Object containing a "template"
+field. The template defines start values and deltas that allow any entry to be
+calculated without explicitly listing each item.
+
+Entry values are calculated as follows, where n is the zero-based entry index:
+
+~~~
+mediaTime[n] = startMediaTime + (n * deltaMediaTime)
+location[n] = [startLocation[0] + (n * deltaLocation[0]),
+               startLocation[1] + (n * deltaLocation[1])]
+wallclock[n] = startWallclock + (n * deltaWallclock)
+~~~
+
+An example media timeline using the template format is shown below:
+
+~~~json
+{
+  "template": {
+    "startMediaTime": 0,
+    "deltaMediaTime": 2002,
+    "startLocation": [0, 0],
+    "deltaLocation": [1, 0],
+    "startWallclock": 1759924158381,
+    "deltaWallclock": 2002,
+    "count": 5
+  }
+}
+~~~
+
+This template represents the same timeline as the explicit entry example above.
+
+### Hybrid format {#hybridformat}
+Publishers MAY combine templates with explicit entries. When both are present,
+the explicit entries in the "entries" array represent additional records that
+extend beyond the template range. Explicit entries are appended after the
+template-generated entries.
+
+~~~json
+{
+  "template": {
+    "startMediaTime": 0,
+    "deltaMediaTime": 2002,
+    "startLocation": [0, 0],
+    "deltaLocation": [1, 0],
+    "startWallclock": 1759924158381,
+    "deltaWallclock": 2002,
+    "count": 100
+  },
+  "entries": [
+    [200200, [100, 0], 1759924358381],
+    [203000, [101, 0], 1759924361181]
+  ]
+}
+~~~
+
+In this example, entries 0-99 are calculated from the template, and entries
+100-101 are provided explicitly with non-constant intervals.
 
 ## Media Timeline Catalog requirements
 A media timeline track MUST carry a 'type' identifier in the Catalog with a value
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1033,6 +1033,30 @@ synchronized data.
 
 ~~~
 
+### Media timeline with template format
+
+This example shows a media timeline track payload using a template-based
+format for constant duration GOPs. This format significantly reduces payload
+size compared to explicit entries.
+
+~~~json
+{
+  "template": {
+    "startMediaTime": 0,
+    "deltaMediaTime": 2002,
+    "startLocation": [0, 0],
+    "deltaLocation": [1, 0],
+    "startWallclock": 1759924158381,
+    "deltaWallclock": 2002,
+    "count": 1800
+  }
+}
+~~~
+
+This single template object represents 1 hour of content with 2-second GOPs
+(1800 entries), replacing what would otherwise be an array of 1800 explicit
+entries.
+
 ### Terminating a live broadcast
 
 This example shows catalog for a media producer terminating a previously
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -227,6 +227,13 @@ Table 1 provides an overview of all fields defined by this document.
 | Language                | lang                   | {{language}}              |
 | Parent name             | parentName             | {{parentname}}            |
 | Track duration          | trackDuration          | {{trackduration}}         |
+| Start media time        | startMediaTime         | {{startmediatime}}        |
+| Delta media time        | deltaMediaTime         | {{deltamediatime}}        |
+| Start location          | startLocation          | {{startlocation}}         |
+| Delta location          | deltaLocation          | {{deltalocation}}         |
+| Start wallclock         | startWallclock         | {{startwallclock}}        |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -227,6 +227,13 @@ Table 1 provides an overview of all fields defined by this document.
 | Language                | lang                   | {{language}}              |
 | Parent name             | parentName             | {{parentname}}            |
 | Track duration          | trackDuration          | {{trackduration}}         |
+| Start media time        | startMediaTime         | {{startmediatime}}        |
+| Delta media time        | deltaMediaTime         | {{deltamediatime}}        |
+| Start location          | startLocation          | {{startlocation}}         |
+| Delta location          | deltaLocation          | {{deltalocation}}         |
+| Start wallclock         | startWallclock         | {{startwallclock}}        |
+| Delta wallclock         | deltaWallclock         | {{deltawallclock}}        |
+| Entry count             | count                  | {{entrycount}}            |
```
wilaw: These should not be defined in this table, as you are not using them directly in the catalog. Instead they should be defined as KVP in the JSON definition of the MediaTimeline section of the spec.

draft-ietf-moq-msf.md
```diff
@@ -517,6 +523,57 @@ Location: T    Required: Optional   JSON Type: Number
...
+The constant interval between wallclock times in a templated media timeline,
+expressed in milliseconds. This field MUST only appear within a template object
+in a media timeline track payload.
+
+### Entry count {#entrycount}
+Location: T    Required: Optional   JSON Type: Number
+
+The number of entries represented by a template in a templated media timeline.
+This field MUST only appear within a template object in a media timeline track
+payload.
```
wilaw: To be meaningful, this field needs a reference timestamp at which this count was accurate.

draft-ietf-moq-msf.md
```diff
@@ -1047,6 +1132,66 @@ An example media timeline is shown below:
...
+  "entries": [
+    [200200, [100, 0], 1759924358381],
+    [203000, [101, 0], 1759924361181]
+  ]
+}
+~~~
+
+In this example, entries 0-99 are calculated from the template, and entries
+100-101 are provided explicitly with non-constant intervals.
+
```
wilaw: I think we should remove Hybrid format for now. It complicates the implementation. A template is used for highly regular segment production. The explicit entry format can be used for everything else. We can add it in later if there is a strong implementation use-case.

draft-ietf-moq-msf.md
wilaw: We can omit this entry count field. The whole point of a template is that it is inviolate and need only be transmitted once.

draft-ietf-moq-msf.md
```diff
@@ -976,9 +1033,33 @@ synchronized data.
...
+{
+  "template": {
+    "startMediaTime": 0,
+    "deltaMediaTime": 2002,
+    "startLocation": [0, 0],
+    "deltaLocation": [1, 0],
+    "startWallclock": 1759924158381,
+    "deltaWallclock": 2002,
+    "count": 1800
+  }
```
wilaw: I think these atteibutes should be in the MediaTimeline Track entry in the catalog. There is no need to create a separate track to transmit this informaiton.

wilaw (commented):
Thanks for your work on this. My primary feedback is that the template version of the mediatimeline can simply be defined as a new packaging type, to differentiate it form the listtype. For example, a mediatimeline entry in the catalog could look like this:

~~~json
{
  "version": 1,
  "generatedAt": 1746104606044,
  "tracks": [
    {
      "name": "history",
      "namespace": "conference.example.com/conference123/alice",
      "packaging": "mediatimelinetemplate",
      "mimetype": "application/json",
      "depends": ["1080p-video","audio"]
    },
~~~

and then the mediatimeline payload could look like this

~~~json
{
    "startMediaTime": 0,
    "deltaMediaTime": 2002,
    "startLocation": [0, 0],
    "deltaLocation": [1, 0],
    "startWallclock": 1759924158381,
    "deltaWallclock": 2002
}
~~~


I also suggest we can remove hybrid mediatimelines for now for simplicity. If we need them in the future, we can just add them as a new packaging type.

suhasHere:
@wilaw please check again and see if the new structure looks about right ? Thanks

draft-ietf-moq-msf.md
```diff
@@ -976,9 +976,61 @@ synchronized data.
...
+      "packaging": "loc",
+      "isLive": true,
+      "targetLatency": 2000,
+      "role": "video",
+      "renderGroup": 1,
+      "codec":"av01.0.08M.10.0.110.09",
+      "width":1920,
+      "height":1080,
+      "framerate":30,
+      "bitrate":1500000
```
wilaw, line 1018: Since the template does not (cannot) change once publishing starts, why don;t we add it directly to the track definition? This is more efficient than the player having to subscribe to a resource just to read a JSON block of constants. We could pack everything into a single new track attribute called "template". Fro example:

```json
{
      "name": "1080p-video",
      "namespace": "conference.example.com/conference123/alice",
      "packaging": "loc",
      "isLive": true,
      "targetLatency": 2000,
      "role": "video",
      "renderGroup": 1,
      "codec":"av01.0.08M.10.0.110.09",
      "width":1920,
      "height":1080,
      "framerate":30,
      "bitrate":1500000,
      "template" : [0,2002,[0,0],[1,0],1759924158381,2002]
}
```

draft-ietf-moq-msf.md
```diff
@@ -1059,6 +1115,73 @@ of each MOQT Group of a media timeline track. The publisher MAY publish incremen
...
+explicit media timeline track. The template approach is best suited for content with
+fixed-duration segments, such as VOD assets or live broadcasts with constant segment
+durations. For content with variable segment durations, the explicit media timeline
+track ({{mediatimelinepayload}}) SHOULD be used instead.
+
+### Media Timeline Template payload {#mediatimelinetemplatepayload}
+A media timeline template track is a JSON {{JSON}} document. This document MAY be
+compressed using GZIP {{GZIP}}. The document contains a JSON Object with the
+following required fields:
+
```
wilaw: We can define a new attribute for a track instead of a new media timeline payload.

wilaw (commented):
See my comments around replacing the mediatimelinetemplate payload with a single new attribute for a track. This would be more efficient for the player to retrieve (no subscription required), would allow audio and video tracks to have different group durations and would be 1xRTT faster for the player to read.

suhasHere:
@wilaw latest changes should make template inline. Thanks

draft-ietf-moq-msf.md
```diff
@@ -1005,23 +1066,27 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
 relationship to wallclock time and media time. Media timeline tracks allow players to
 seek to precise points behind the live head in a live broadcast, or for random access
 in a VOD asset. Media timeline tracks are optional. Multiple media timeline tracks
 can exist inside a catalog.
 
 ## Media Timeline track payload {#mediatimelinepayload}
 A media timeline track is a JSON {{JSON}} document. This document MAY be compressed
-using GZIP {{GZIP}}. The document contains an array of records. Each record consists of
+using GZIP {{GZIP}}. The document supports two formats: an explicit entry format
+and a template format. Publishers MAY combine both formats in a single document.
```
wilaw, line 1086: There are not two 'formats' for the media timeline track, only one. I would remove the refernece to 'explicit entires' and in fact leave the media timeline secitonb as it was before. 

I would not specify the media timeline template inside this section, which is the Media Timeline track payload , as the template definition is is never found inside the media timeline track payload. 

Instead, the templater definition can just be moved to a separate template section, in the section where all the other track properties are defined.

draft-ietf-moq-msf.md
```diff
@@ -1005,23 +1066,27 @@ group, represented as the number of milliseconds since the Unix epoch, rounded t
...
+
+### Explicit entry format {#explicitentryformat}
```
wilaw, line 1088: Remove this reference to "explciti entry format". . There is only one way encoding a media timeline payload.

draft-ietf-moq-msf.md
```diff
@@ -1030,12 +1095,12 @@ an array of three required items, whose ordinal position defines their type:
...
   Group ID and MOQT Object ID, and expressed as a JSON Array of Numbers, where the
   first number is the Group ID and the second number is the Object ID.
-* The third time holds the wallclock time at which the media was encoded, defined as
+* The third item holds the wallclock time at which the media was encoded, defined as
   the number of milliseconds that have elapsed since January 1, 1970
   (midnight UTC/GMT) and expressed as a JSON Number. For VOD assets, or if the
   wallclock time is not known, the value SHOULD be 0.
 
-An example media timeline is shown below:
+An example media timeline using explicit entries is shown below:
```
```suggestion
An example media timeline is shown below:
```

draft-ietf-moq-msf.md
```diff
@@ -1059,6 +1124,62 @@ of each MOQT Group of a media timeline track. The publisher MAY publish incremen
 updates in the second and subsequent Objects within each Group. Incremental updates
 only contain media timeline records since the last media timeline Object.
 
+## Media Timeline Template {#mediatimelinetemplate}
```
wilaw, line 1127: Move this whole section to your ### Template {#template} definition.

draft-ietf-moq-msf.md
```diff
@@ -1059,6 +1124,62 @@ of each MOQT Group of a media timeline track. The publisher MAY publish incremen
...
+When the relationship between media time, MOQT Location and wallclock time follows
+a regular, predictable pattern, a media timeline template MAY be used instead of an
+explicit media timeline track. The template approach is best suited for content with
+fixed-duration segments, such as VOD assets or live broadcasts with constant segment
+durations. For content with variable segment durations, the explicit media timeline
+track ({{mediatimelinepayload}}) SHOULD be used instead.
+
+### Inline Template Format
```
```suggestion
### Template Format
```

draft-ietf-moq-msf.md
```diff
@@ -1059,6 +1124,62 @@ of each MOQT Group of a media timeline track. The publisher MAY publish incremen
...
+   timeline entries in {{explicitentryformat}}. For VOD assets, or if the wallclock
+   time is not known, the value SHOULD be 0.
+6. deltaWallclock - The constant interval between wallclock times, expressed as a
+   JSON Number in milliseconds. For VOD assets, or if the wallclock time is not
+   known, the value SHOULD be 0.
+
+All six values are mandatory and MUST appear in the specified order.
+
+Clients compute entry values using the following formulas, where n is the
+zero-based entry index:
```
wilaw, line 1159: where n is set to zero to calculate the initial entry and raised by one for each subsequent entry.

wilaw (changes requested):
The template definition is nicely done, but it should be moved out of the media timeline payload section, and into the section which defines the track property "template".

draft-ietf-moq-msf.md
```diff
@@ -1059,6 +1124,62 @@ of each MOQT Group of a media timeline track. The publisher MAY publish incremen
 updates in the second and subsequent Objects within each Group. Incremental updates
 only contain media timeline records since the last media timeline Object.
 
+## Media Timeline Template {#mediatimelinetemplate}
+When the relationship between media time, MOQT Location and wallclock time follows
+a regular, predictable pattern, a media timeline template MAY be used instead of an
+explicit media timeline track. The template approach is best suited for content with
+fixed-duration segments, such as VOD assets or live broadcasts with constant segment
+durations. For content with variable segment durations, the explicit media timeline
+track ({{mediatimelinepayload}}) SHOULD be used instead.
```
```suggestion
track ({{mediatimelinepayload}}) MUST be used instead.
```

wilaw (approved):
I'd like to see the template definition be independent of the mediatime track payload definition, since the two are quite different - one is a track property in the catalog, the other is a payload within an object of a track. However this is a hard change to make in this PR. Let's merged this, and then reformat in a subsequent PR.

PR #132, opened 2026-02-26
Allow targetLatency when isLive is FALSE

### LINKED ISSUE #98: Allow targetLatency present when isLive is FALSE?
It is currently specified that the targetLatency field MUST NOT be present when isLive is FALSE. However, since it is possible to turn a previously live content to be a VOD content, this requirement would then require the removal of the targetLatency field from each of all tracks when turning a previously live content to be a VOD content.

It seems to me that it'd be better to allow the presence of the targetLatency field when isLive is FALSE, but just ignore it, when present, if isLive is FALSE.

suhasHere:
Fixes #98

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -379,11 +379,11 @@ The target latency in milliseconds. Target latency is defined as the offset in
 wallclock time between when content was encoded and when it is displayed to the
 end user. For example, if a frame of video is encoded at 10:08:32.638 UTC and the
 target latency is 5000, then that frame should be rendered to the end-user at
-10:08:37.638 UTC. This field MUST NOT be included if isLive is FALSE. All tracks
+10:08:37.638 UTC. If isLive is FALSE, this field MUST be ignored. All tracks
 belonging to the same render group MUST have identical target latencies. All tracks
 belonging to the same alternate group MUST have identical target latencies. If this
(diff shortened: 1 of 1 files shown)


wilaw (approved):

draft-ietf-moq-msf.md
```diff
@@ -379,11 +379,11 @@ The target latency in milliseconds. Target latency is defined as the offset in
 wallclock time between when content was encoded and when it is displayed to the
 end user. For example, if a frame of video is encoded at 10:08:32.638 UTC and the
 target latency is 5000, then that frame should be rendered to the end-user at
-10:08:37.638 UTC. This field MUST NOT be included if isLive is FALSE. All tracks
+10:08:37.638 UTC. If isLive is FALSE, this field MUST be ignored. All tracks
 belonging to the same render group MUST have identical target latencies. All tracks
 belonging to the same alternate group MUST have identical target latencies. If this
-field is absent from the track definition, then the player MAY choose the latency
-with which it renders the content.
+field is absent from the track definition, or if isLive is FALSE, then the player
```
```suggestion
field is absent from the track definition, and isLive is TRUE, then the player
```
We already define that if isLive is FALSE this field can be ignored, so we only need to clarify that if it is not present and the track is live, that the plater can choose its own target latency.

wilaw (approved):

PR #141, opened 2026-02-28
Add support for InitTracks

### LINKED ISSUE #138: `initData` should use separate tracks
The current scheme is simple, but has some problems:
1. Players are forced to download init segments for all renditions.
2. It's not possible to update the init segment, ex. a new group when the SPS/PPS changes.
3. There's no way to reuse init data across tracks, even if they're duplicate.
4. Updating tracks via add/remove means repeating the initData.
5. Base64 is less efficient.

Yes, putting the init segment in the catalog can avoid a RTT when using CMAF. But I agree with @mzanaty that it's a flaw with CMAF for not inlining the damn stuff. And if that RTT is really important... `PUBLISH` the init tracks to get the current behavior.

I also have an extra use-case that won't work unless init segments are separate tracks: I want to serve objects via HTTP for HLS/DASH backwards compatibility. That's not possible to do in a generic way unless init data is a separate track.

suhasHere:
Fixes #138

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -424,18 +424,9 @@ Location: T    Required: Optional   JSON Type: String
 
 A string specifying the name of a track that holds the initialization data
 for this track. The referenced track MUST exist in the same catalog and share
-the same namespace as the referencing track. The initialization track publishes
-raw (non-Base64 encoded) initialization segment data as its payload.
-
-Using a separate initialization track instead of inline initData {{initdata}}
-enables:
-
-* Independent updates to initialization data without modifying the catalog
-* Deduplication when multiple tracks share identical initialization segments
-* HTTP-addressable initialization segments for compatibility with HLS/DASH
-  delivery
-
-A track MUST NOT specify both initTrack and initData.
+the same namespace as the referencing track. A track MUST NOT specify both
+initTrack and initData. See {{initializationtrack}} for initialization track
+requirements and payload format.
 
 ### Dependencies {#dependencies}
 Location: T    Required: Optional   JSON Type: Array
@@ -1102,7 +1093,7 @@ The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
 of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
 MOQT Stream.
 
-## Group numbering
+## Group numbering {#group-numbering}
 The Group ID of the first Group published in a track at application startup MUST be
 a unique integer that will not repeat in the future. One approach to achieve this
 is to set the initial Group ID to the creation time of the first Object in the
@@ -1135,9 +1126,8 @@ The initialization track payload contains raw (non-Base64 encoded) codec
 initialization data. The format of this data depends on the media packaging
 {{mediapackaging}} and codec in use:
 
-* For LOC packaging with ISOBMFF-based codecs (H.264, H.265, AV1, AAC), the payload
-  contains the concatenation of the ftyp and moov boxes that form the initialization
-  segment.
+* For LOC packaging, the payload contains codec-specific initialization data as
+  defined in {{LOC}}.
 * For other container formats, the payload contains the codec-specific initialization
   data as defined by that format.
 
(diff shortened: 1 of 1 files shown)


vasilvv:
In this design, how do I sync new init segments against specific points in the track?

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,73 @@ Each subsequent Group ID MUST increase by 1.
...
+
+An initialization track provides codec initialization data for one or more media
+tracks. Instead of embedding initialization data directly in the catalog using the
+initData field {{initdata}}, publishers can reference a separate initialization track
+using the initTrack field {{inittrack}}. This approach offers several advantages:
+
+* Independent updates to initialization data without modifying the catalog
+* Deduplication when multiple tracks share identical initialization segments
+* HTTP-addressable initialization segments for compatibility with HLS/DASH delivery
+* Reduced catalog size for codecs with large initialization data
```
wilaw, line 1121: We don't need to list the advantages in the spec, only the implementation.

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,73 @@ Each subsequent Group ID MUST increase by 1.
...
+# Initialization track {#initializationtrack}
+
+An initialization track provides codec initialization data for one or more media
+tracks. Instead of embedding initialization data directly in the catalog using the
+initData field {{initdata}}, publishers can reference a separate initialization track
+using the initTrack field {{inittrack}}. This approach offers several advantages:
+
+* Independent updates to initialization data without modifying the catalog
+* Deduplication when multiple tracks share identical initialization segments
+* HTTP-addressable initialization segments for compatibility with HLS/DASH delivery
```
```suggestion
* Compatibility with HLS/DASH delivery
```

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,73 @@ Each subsequent Group ID MUST increase by 1.
...
+
+## Initialization track group and object structure
+
+An initialization track publishes its payload as a single MOQT Object within a
+MOQT Group. The Object ID MUST be 0.
+
+When the initialization data changes (for example, due to a codec parameter change
+or resolution switch that requires new decoder configuration), the publisher MUST
+publish a new Group with an incremented Group ID containing the updated
+initialization data.
```
```suggestion
An initialization track initially publishes its payload as a single MOQT Object within a
MOQT Group. The Object ID MUST be 0. The Group ID MUST match the starting Group ID
of the tracks which reference this initialization track in their initTrack property

If the initialization data then changes (for example, due to a codec parameter change
or resolution switch that requires new decoder configuration), the publisher MUST
publish a new Group with an incremented Group ID containing the updated
initialization data. That new Group ID MUST again match the target Group ID at which the
new initialization data is required by the tracks which reference this initialization track
in their initTrack property. 

```

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,73 @@ Each subsequent Group ID MUST increase by 1.
...
+A track that specifies an initTrack field MUST NOT also specify an initData field.
+The referenced initialization track MUST exist in the same catalog.
+
+## Initialization track updating
+
+For live streams, the publisher SHOULD publish the initialization data in
+Object 0 of the first Group at the start of the broadcast. The first Group ID
+MUST follow the requirements in {{group-numbering}}. If the initialization data
+needs to change during the broadcast (such as a mid-stream resolution change),
+the publisher increments the Group ID and publishes the new initialization data.
```
```suggestion
For live streams, the publisher MUST publish the initialization data in
Object 0 of the first Group at the start of the broadcast. The first Group ID
MUST follow the requirements in {{group-numbering}} and MUST match the 
Group ID of the tracks which uses this initTrack for initialization.  If the initialization data
needs to change during the broadcast (such as a mid-stream resolution change),
the publisher increments the Group ID to match the point at which the change applies in the
dependent tracks and publishes the new initialization data.
```

wilaw (commented):
If you agree with these changes for signaling where a mid-stream init change should apply, I'll merge this PR. If not, then can you propose an alternate?

suhasHere:
@wilaw  I agree with your proposal. Let's get this landed and I do have few editorial nits, but that can come later.

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,79 @@ Each subsequent Group ID MUST increase by 1.
...
+The initialization data MUST be sufficient to initialize a decoder for all tracks
+that reference this initialization track.
+
+## Initialization track group and object structure
+
+An initialization track initially publishes its payload as a single MOQT Object within a
+MOQT Group. The Object ID MUST be 0. The Group ID MUST match the starting Group ID
+of the tracks which reference this initialization track in their initTrack property
+
+If the initialization data then changes (for example, due to a codec parameter change
```
```suggestion
of the tracks which reference this initialization track in their initTrack property.

If the initialization data then changes (for example, due to a codec parameter
change
```

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,79 @@ Each subsequent Group ID MUST increase by 1.
...
+
+A track that specifies an initTrack field MUST NOT also specify an initData field.
+The referenced initialization track MUST exist in the same catalog.
+
+## Initialization track updating
+
+For live streams, the publisher MUST publish the initialization data in
+Object 0 of the first Group at the start of the broadcast. The first Group ID
+MUST follow the requirements in {{group-numbering}} and MUST match the 
+Group ID of the tracks which uses this initTrack for initialization.  If the initialization data
```
```suggestion
Object 0 of the first Group at the start of the broadcast. The first Group ID
MUST follow the requirements in {{group-numbering}} and MUST match the
Group ID of the tracks which uses this initTrack for initialization. If the
initialization data
```

draft-ietf-moq-msf.md
```diff
@@ -1018,6 +1108,81 @@ Each subsequent Group ID MUST increase by 1.
...
+If the initialization data then changes (for example, due to a codec parameter
+change
+or resolution switch that requires new decoder configuration), the publisher MUST
+publish a new Group with an incremented Group ID containing the updated
+initialization data. That new Group ID MUST again match the target Group ID at which the
+new initialization data is required by the tracks which reference this initialization track
+in their initTrack property. 
+
+
+Subscribers can either fetch or subscribe to the initialization track. When
```
wilaw, line 1150:
```suggestion
new initialization data is required by the tracks which reference this initialization track
in their initTrack property.

Subscribers can either fetch or subscribe to the initialization track. When
```

PR #143, opened 2026-03-01
break the monolith table into separate tables and sections

### LINKED ISSUE #137: Editorial: Split out Root and Track schemas
The draft puts all catalog fields in one section, with a note indicating if the fields are either in the root or track. However, none of the fields are currently in both, so it's really confusing to flatten them together.

Can you make two sections, one for fields in the root and one for fields in the track?

suhasHere:
Fixes #137 

This is purely editorial, but does move a few things around.

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -197,7 +197,7 @@ Table 1 lists the fields defined at the root of the catalog JSON object.
 | Tracks                  | tracks                 | {{tracks}}                |
 
 ### MSF version {#msfversion}
-Required: Yes    JSON Type: Number    Location: Root Catalog (Table 1)
+Required: Yes    JSON Type: Number    Location: Root Catalog
 
 Specifies the version of MSF referenced by this catalog. There is no guarantee
 that future catalog versions are backwards compatible and field definitions and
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -187,171 +185,173 @@ of field names within the JSON document is not important.
...
 ### Tracks object {#trackobject}
 
-A track object is JSON Object containing a collection of fields whose location
-is specified 'T' in Table 2.
+A track object is a JSON Object containing a collection of fields whose location
+is specified in Table 3.
 
 ### Track namespace {#tracknamespace}
-Location: T    Required: Optional    JSON Type: String
+Required: Optional    JSON Type: String    Location: Track Object (Table 3)
```
wilaw: We don't need to repeat Track Object (Table 3) 15 times. Since you have labelled Table 3 as 'Track Object Fields', this is clear to the reader, and the text can be simplifed to just say 'Track Object'. 

This comment applies to all the track field definitions. 

```suggestion
Required: Optional    JSON Type: String    Location: Track Object
```

draft-ietf-moq-msf.md
```diff
@@ -187,171 +185,173 @@ of field names within the JSON document is not important.
...
-
-| Location |                Allowed locations for the field                |
-|:=========|:==============================================================|
-| R        | The Root of the JSON object                                   |
-| T        | Track object                                                  |
-
 
 ### MSF version {#msfversion}
-Location: R    Required: Yes    JSON Type: Number
+Required: Yes    JSON Type: Number    Location: Root Catalog (Table 1)
```
wilaw: Don't need to add the table number to every Location: Root Catalog entry. It is already clear enough. 

```suggestion
Required: Yes    JSON Type: Number    Location: Root Catalog
```

wilaw (changes requested):
I think this separation is nice and clear. However I think we don't need to repeat the table number in every root or track property definition. This is repetitive information which is also brittle - if we insert a table in the future we need to roll every property definition.

suhasHere:
@wilaw does the new changes look good ? Thanks

wilaw (approved):
Those changes were spot-on. This is a nice improvement to the clarity of the doc.

PR #152, opened 2026-04-03
Clarify MSF URL construction and fragment parameters

### LINKED ISSUE #151: Update URl syntax to match MOQT
MOQT now has defined a URI syntax https://moq-wg.github.io/moq-transport/draft-ietf-moq-transport.html#name-moqt-uri-scheme 

moqt://example.com/app#<type>:<value>

We need to ensure that MSF extends that and also that we register "msf" as a reserved type.

wilaw:
Updated the MSF URL construction and interpretation section, clarifying the msf-fragment and its parameters, including connection types and examples.

Fixes: #151

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1568,49 +1568,86 @@ This example shows drone GPS coordinates synched with the start of each Group.
 ## URL construction and interpretation
 
 An MSF URL identifies a MOQT session and an optional sub-resource within that session.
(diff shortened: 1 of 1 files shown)


gwendalsimon (approved):
I think it is correct. Thank you for putting this in a consistent manner. I understand that, when a range is indicated in the fragment, then it is a FETCH and when no range is given, I would expect a SUBSCRIBE to be issued.

Issue #148, opened 2026-03-02
Media Mapping to MoQT


kixelated:
> [6. ](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#section-6)[Media transmission](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#name-media-transmission)
The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective of the [Section 4](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#mediapackaging) in place, each MOQT Object MUST be mapped to a new MOQT Stream.

If I'm reading this correctly, MSF requires one stream/sub-group per object? If that's intentional, it's extremely restrictive. 

Most people intend to use sub-group 0 only, or a sub-group per SVC layer. A QUIC/MoQT stream per frame/object should definitely not be a requirement.

wilaw:
We can restrict to subgroup 0 only for now. If someone defines a SVC capability later, we can modify that.

We follow [LOC packaging](https://datatracker.ietf.org/doc/html/draft-mzanaty-moq-loc-05), which states that "_The LOC Payload is the "internal data" of an EncodedAudioChunk or EncodedVideoChunk._" restricts packaging .  If that's overlay restrictive - then you need to lobby Mo to get it changed within the LOC draft. 

With CMSF packaging (in CMSF), we allow at least one chunk per stream, and at most one Group group per stream.

suhasHere:
On stream per object is definitely not the goal with LOC. Actually LOC shouldn't care about stream mapping. This needs to be fixed. I will check on Loc side to make sure its the case and open issue, if that's true.

In our implementation, we encode, package into Loc and send on a new stream only if it is a new group

PR #156, opened 2026-05-11
Make MOQT Object to Stream mapping implementation-specific

suhasHere:
@wilaw @kixelated does this address #148

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1568,9 +1568,13 @@ In this example:
   as defined in the respective specifications.
 
 # Media transmission
-The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
-of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
-MOQT Stream.
+The {{mediapackaging}} defines how media samples are packaged into MOQT Objects.
+The mapping of MOQT Objects to MOQT Streams is implementation-specific. Examples
+of stream mapping strategies include:
+
+* Mapping each MOQT Object to a separate MOQT Stream
+* Mapping all Objects within a subgroup to a single MOQT Stream
+* Mapping Objects by temporal or spatial layer to separate MOQT Streams
 
 ## Group numbering
 The Group ID of the first Group published in a track at application startup MUST be
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -381,6 +381,7 @@ Table 3 lists the fields defined within each track object.
 | Parent name             | parentName             | {{parentname}}            |
 | Track duration          | trackDuration          | {{trackduration}}         |
 | Authorization Info      | authInfo               | {{authinfo}}              |
+| Stream mapping          | streamMapping          | {{streammapping}}         |
 
 ### Tracks object {#trackobject}
 
@@ -760,6 +761,30 @@ A catalog with:
 
 Would be resolved by the subscriber to include `"cat": "XYZ789"`, which is
 then presented in control messages as specified by the authorization scheme.
+
+### Stream mapping {#streammapping}
+Required: Optional    JSON Type: String    Location: Track Object
+
+A string defining how MOQT Objects in this track are mapped to MOQT Streams.
+When this field is absent, the default mapping is "stream-per-group".
+Allowed values are defined in Table 8.
+
+| Name                |  Value                |  Description                                          |
+|:====================|:======================|:======================================================|
+| Stream per object   | stream-per-object     | Each MOQT Object is mapped to a separate MOQT Stream |
+| Stream per subgroup | stream-per-subgroup   | All Objects within a Subgroup share a single MOQT Stream |
+| Stream per group    | stream-per-group      | All Objects within a Group share a single MOQT Stream |
+
+Table 8: Allowed stream mapping values
+
+All stream mapping values in Table 8 are valid for LOC packaged tracks. Other
+packaging specifications MAY define additional stream mapping values or restrict
+the set of allowed values for their packaging type.
+
+The choice of stream mapping affects latency and head-of-line blocking
+characteristics. For example, "stream-per-object" minimizes head-of-line blocking
+at the cost of more QUIC streams, while "stream-per-group" reduces stream overhead
+but may increase latency under packet loss.
 
 ## Delta updates {#deltaupdates}
 A catalog update might contain incremental changes. This is a useful property if
@@ -1569,12 +1594,13 @@ In this example:
 
 # Media transmission
 The {{mediapackaging}} defines how media samples are packaged into MOQT Objects.
-The mapping of MOQT Objects to MOQT Streams is implementation-specific. Examples
-of stream mapping strategies include:
+For LOC packaged tracks, each encoded media chunk (EncodedAudioChunk or
+EncodedVideoChunk) is mapped to a single MOQT Object as defined in {{mediapackaging}}.
 
-* Mapping each MOQT Object to a separate MOQT Stream
-* Mapping all Objects within a subgroup to a single MOQT Stream
-* Mapping Objects by temporal or spatial layer to separate MOQT Streams
+The mapping of MOQT Objects to MOQT Streams is signaled per-track in the catalog
+using the stream mapping field {{streammapping}}. Publishers MUST include the
+streamMapping field in the catalog for each media track. Subscribers MUST use
+the signaled stream mapping when consuming the track.
 
 ## Group numbering
 The Group ID of the first Group published in a track at application startup MUST be
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -429,7 +429,7 @@ Table 2 lists the fields defined within each track object.
 | Parent namespace        | parentNamespace        | {{parentnamespace}}       |
 | Track duration          | trackDuration          | {{trackduration}}         |
 | Authorization Info      | authInfo               | {{authinfo}}              |
-| Stream mapping          | streamMapping          | {{streammapping}}         |
+| Subgroup mapping        | subgroupMapping        | {{subgroupmapping}}       |
 | Accessibility           | accessibility          | {{accessibility}}         |
 
 ### Tracks object {#trackobject}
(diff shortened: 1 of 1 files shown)


wilaw (commented):
I think this change is under defined. Saying "The mapping of MOQT Objects to MOQT Streams is implementation-specific." leaves both publishers and subscribers developers wondering what exactly they should produce or consume. We are defining the implementation and therefore we should define the types of mapping that are allowed and also communicate them. We could define a table of allowed mapping types and then for each track definition, communicate the type of mapping that it deploys. The table needs to define how we map media to MOQT Groups and Objects, and then how those Groups and Objects are mapped to QUIC streams and datagrams. 


| Type | Media to MOQT | MOQT to QUIC |
| ----- | ----------- | ---------|
| 1        | Each video frame in a separate Object and each GOP in a separate Group | Each MOQT Object in a separate MOQT stream  |
| 2        | Each video frame in a separate Object and each GOP in a separate Group | Each MOQT Group in a separate MOQT stream  |
| 3       | Each video GOP in a separate Object and one object per Group|  Each MOQT Object in a separate MOQT stream  |
| 4       | Each audio sample in a separate Object and each GOP in a separate Group | Each MOQT Object in a separate MOQT stream  |

... etc

kixelated:
Yeah I think it would be a mistake to have it be undefined. 

But the container should probably define the mapping. CMSF would be a GoP per group and fragment per object, for example.

suhasHere:
> I think this change is under defined. Saying "The mapping of MOQT Objects to MOQT Streams is implementation-specific." leaves both publishers and subscribers developers wondering what exactly they should produce or consume. We are defining the implementation and therefore we should define the types of mapping that are allowed and also communicate them. We could define a table of allowed mapping types and then for each track definition, communicate the type of mapping that it deploys. The table needs to define how we map media to MOQT Groups and Objects, and then how those Groups and Objects are mapped to QUIC streams and datagrams. 
> 
> 
> | Type | Media to MOQT | MOQT to QUIC |
> | ----- | ----------- | ---------|
> | 1        | Each video frame in a separate Object and each GOP in a separate Group | Each MOQT Object in a separate MOQT stream  |
> | 2        | Each video frame in a separate Object and each GOP in a separate Group | Each MOQT Group in a separate MOQT stream  |
> | 3       | Each video GOP in a separate Object and one object per Group|  Each MOQT Object in a separate MOQT stream  |
> | 4       | Each audio sample in a separate Object and each GOP in a separate Group | Each MOQT Object in a separate MOQT stream  |
> 
> ... etc


Makes sense Will. This reminded me of this document I help write a while ago ( https://datatracker.ietf.org/doc/html/draft-jennings-moq-usages-00), more specifically : https://datatracker.ietf.org/doc/html/draft-jennings-moq-usages-00#section-9

I will work on updating the Pr along the lines of your suggestions here

draft-ietf-moq-msf.md
```diff
@@ -835,30 +762,29 @@ A catalog with:
...
+All stream mapping values in Table 8 are valid for LOC packaged tracks. Other
+packaging specifications MAY define additional stream mapping values or restrict
+the set of allowed values for their packaging type.
 
-A subscriber MAY use this information to determine caption availability
-and configure an appropriate caption decoder.
+The choice of stream mapping affects latency and head-of-line blocking
+characteristics. For example, "stream-per-object" minimizes head-of-line blocking
+at the cost of more QUIC streams, while "stream-per-group" reduces stream overhead
+but may increase latency under packet loss.
```
wilaw: We should not specify what is valid for LOC packaging, the LOC spec can do that. This section can be removed. We also don't need to motivate the reasons for one type of packaging over another, that is also the province of the packaging specifications. 

```suggestion

```

draft-ietf-moq-msf.md
```diff
@@ -835,30 +762,29 @@ A catalog with:
...
-
-An array of accessibility descriptors indicating accessibility features
-embedded within the track. Each descriptor is a JSON Object containing:
+### Stream mapping {#streammapping}
+Required: Optional    JSON Type: String    Location: Track Object
 
-* A required 'scheme' field (String) identifying the accessibility scheme.
-* A required 'value' field (String) specifying the accessibility channels
-  or features available.
+A string defining how MOQT Objects in this track are mapped to MOQT Streams.
```
wilaw: Shouldn't  this be QUIC Streams? MOQT does not define streams.

draft-ietf-moq-msf.md
```diff
@@ -835,30 +762,29 @@ A catalog with:
...
+When this field is absent, the default mapping is "stream-per-group".
+Allowed values are defined in Table 8.
 
-Table 6: Registered accessibility schemes
+| Name                |  Value                |  Description                                          |
+|:====================|:======================|:======================================================|
+| Stream per object   | stream-per-object     | Each MOQT Object is mapped to a separate MOQT Stream |
+| Stream per subgroup | stream-per-subgroup   | All Objects within a Subgroup share a single MOQT Stream |
+| Stream per group    | stream-per-group      | All Objects within a Group share a single MOQT Stream |
```
wilaw: Same comment about QUIC Streams

draft-ietf-moq-msf.md
```diff
@@ -1778,25 +1593,29 @@ In this example:
...
+For LOC packaged tracks, each encoded media chunk (EncodedAudioChunk or
+EncodedVideoChunk) is mapped to a single MOQT Object as defined in {{mediapackaging}}.
 
-## Group numbering
-Group IDs for a track MUST be unique and MUST increase monotonically. Within a
-continuous publishing session, each subsequent Group ID SHOULD increase by 1.
+The mapping of MOQT Objects to MOQT Streams is signaled per-track in the catalog
+using the stream mapping field {{streammapping}}. Publishers MUST include the
+streamMapping field in the catalog for each media track. Subscribers MUST use
+the signaled stream mapping when consuming the track.
```
wilaw: We don't need to tell subscribers to use the information send in the catalog. That can be assumed. 

```suggestion
using the stream mapping field {{streammapping}}. Publishers MUST include the
streamMapping field in the catalog for each media track.
```

wilaw (commented):
Main concern is that we should be referencing QUIC streams as "stream" is not a MOQT construct. 

Also, I have a concern about layering. Shouldn't the packaging format (be it LOC or CMAF) define the how the Objects are mapped to QUIC streams? If so, then wouldn't they invent their own mechanism for communicating that? Is this intended as a convenience property which can be used by external packaging formats?

gwendalsimon:
Discussed in London: players usually rebuild this by inspecting. Concerns about terminology and confusion. It may be something for each media container (LOC, CMSF). Allow the publisher to express the dependencies.

useful, fix terminology, having a focus on signalling dependency

draft-ietf-moq-msf.md
```diff
@@ -870,6 +871,31 @@ A catalog with:
...
+A string defining how MOQT Objects in this track are organized into MOQT
+Subgroups. This signals to subscribers what content boundaries they can expect
+within each subgroup. When this field is absent, the default mapping is
+"subgroup-per-group". Allowed values are defined in Table 8.
+
+| Name                  |  Value                  |  Description                                                      |
+|:======================|:========================|:==================================================================|
+| Subgroup per object   | subgroup-per-object     | Each MOQT Object is placed in its own Subgroup                    |
+| Subgroup per GOP      | subgroup-per-gop        | All Objects belonging to a single GOP share a Subgroup            |
+| Subgroup per group    | subgroup-per-group      | All Objects within a Group share a single Subgroup                |
```
wilaw, line 886: This table crosses MOQT/media boundaries. "Pbject" and "Group" are MOQT constructs. GOP is a media construct. If you include that, then you should also have a "subgroup per Frame". I don't think you should include any media entries ion this table. It should have only two rows  - "Subgroup per Object" and 'Subgroup per Group".

draft-ietf-moq-msf.md
```diff
@@ -1824,9 +1850,14 @@ In this example:
...
-of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
-MOQT Stream.
+The {{mediapackaging}} defines how media samples are packaged into MOQT Objects.
+The specific mapping of media frames to MOQT Objects is defined by each packaging
+specification (e.g., LOC, CMSF) and is outside the scope of MSF.
+
+The organization of MOQT Objects into MOQT Subgroups is signaled per-track in the
+catalog using the subgroup mapping field {{subgroupmapping}}. Publishers MUST include
+the subgroupMapping field in the catalog for each media track. Subscribers MUST use
+the signaled subgroup mapping when consuming the track.
```
wilaw, line 1860: You indicate that "When this field is absent, the default mapping is"subgroup-per-group". Therefore I don't think a MUST include is necessary. It SHOULD be included if the subgroup mapping is not the default. 

Also, you can remove the line about "Subscribers MUST use the signaled subgroup mapping when consuming the track.". Subscribers can do anything they like and we cannot test or enforce their compliance.

draft-ietf-moq-msf.md
```diff
@@ -870,6 +871,31 @@ A catalog with:
 Would be resolved by the subscriber to include `"cat": "XYZ789"`, which is
 then presented in control messages as specified by the authorization scheme.
 
+### Subgroup mapping {#subgroupmapping}
```
wilaw, line 874: What if I am send video over datagrams? What subgroup mapping is signaled then?

draft-ietf-moq-msf.md
```diff
@@ -870,6 +871,31 @@ A catalog with:
...
+Required: Optional    JSON Type: String    Location: Track Object
+
+A string defining how MOQT Objects in this track are organized into MOQT
+Subgroups. This signals to subscribers what content boundaries they can expect
+within each subgroup. When this field is absent, the default mapping is
+"subgroup-per-group". Allowed values are defined in Table 8.
+
+| Name                  |  Value                  |  Description                                                      |
+|:======================|:========================|:==================================================================|
+| Subgroup per object   | subgroup-per-object     | Each MOQT Object is placed in its own Subgroup                    |
```
kixelated, line 884: please no. The base draft should absolutely not require support for 3 variants.

Stick with a single group/sub-group per GoP for the base draft. If somebody wants to use sub-groups, that's an extension to MSF because it requires *more information* about the media encoding.

For example, we would need to know the mapping of SVC layers to sub-groups. It's impossible to build a decoder otherwise. 

And same for stuff like sending frames as a separate sub-groups. My decoder needs which data is droppable. Please don't make me inspect the codec bitstream to discover if all frame dependencies have been met before trying to decode, any scheme needs to be encoded in the catalog.

kixelated (changes requested):

gwendalsimon:
I suggest discarding this PR in favor of #169. My reasoning is in
https://github.com/moq-wg/msf/pull/169#issuecomment-5024415228: the base draft
should mandate a single Subgroup (ID 0) per Group, and richer Subgroup layouts
such as SVC belong in an extension that also signals frame dependencies.

wilaw:
Will/Suhas sync - closing this PR in favor of each packaging format defining the mapping of payload contents to transport layer. Note, LOC defines only the container. MSF when specifycing LOC stremaing needs to provide details on how to map LOC Groups to MOQT streams.

Issue #147, opened 2026-03-02
Confusing Media Transmission section


kixelated:
> [6.1. ](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#section-6.1)[Group numbering](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#name-group-numbering)
The Group ID of the first Group published in a track at application startup MUST be a unique integer that will not repeat in the future. One approach to achieve this is to set the initial Group ID to the creation time of the first Object in the group, represented as the number of milliseconds since the Unix epoch, rounded to the nearest millisecond. This ensures that republishing the same track in the future, such as after a loss of connectivity or an encoder restart, will not result in smaller or duplicate Group IDs for the same track name. Note that this method does not prevent duplication if more than 1000 groups are published per second.
>
> Each subsequent Group ID MUST increase by 1.
>
> If a publisher is able to maintain state across a republish, it MUST signal the gap in Group IDs using the MOQT Prior Group ID Gap Extension header.

All three of these paragraphs are confused.

- You can't say Group ID **MUST** increase by 1 except when it can't.
- You can't signal a gap extension if you don't remember the previous group.
- You *technically* can't rely on wall clock timestamps to prevent duplicates.

IMO just say duplicate Group IDs are prohibited, which afaik if a requirement of MoqTransport anyway and doesn't actually need to be mentioned. If you want an editorial section on how to handle republishing the same track, go for it, but you also need to elaborate on how to deal with the catalog too.

wilaw:
Fixed with PR #157

PR #157, opened 2026-05-11
Clarify Group numbering requirements for restarts (#147)

suhasHere:
@kixelated @wilaw does this address #147

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1573,19 +1573,16 @@ of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
 MOQT Stream.
 
 ## Group numbering
-The Group ID of the first Group published in a track at application startup MUST be
-a unique integer that will not repeat in the future. One approach to achieve this
-is to set the initial Group ID to the creation time of the first Object in the
-group, represented as the number of milliseconds since the Unix epoch, rounded to
-the nearest millisecond. This ensures that republishing the same track in the
-future, such as after a loss of connectivity or an encoder restart, will not result
-in smaller or duplicate Group IDs for the same track name. Note that this method
-does not prevent duplication if more than 1000 Groups are published per second.
+Group IDs for a track MUST be unique and MUST increase monotonically. Within a
+continuous publishing session, each subsequent Group ID MUST increase by 1.
 
-Each subsequent Group ID MUST increase by 1.
+When a publisher restarts (e.g., after connectivity loss or encoder restart), it
+MUST ensure the new starting Group ID is greater than any previously published
+Group ID for that track. One approach is to use the current wall clock time in
+milliseconds since the Unix epoch as the starting Group ID.
 
-If a publisher is able to maintain state across a republish, it MUST signal the gap
-in Group IDs using the MOQT Prior Group ID Gap Extension header.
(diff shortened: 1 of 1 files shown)


wilaw (approved):
These are reasonable solutions to the issues Luke exposed.

kixelated:
I would remove the "MUST increase by 1" part because it's not enforceable, and contradicts the support for discontinuities. 

CMSF could say SHOULD increase by 1. I'm not sure if MSF should.

suhasHere:
I do understand its no enforceable , but 2 msf clients can infer use this MUST to decipher a loss if there is a gap.

wilaw:
> I would remove the "MUST increase by 1" part because it's not enforceable, and contradicts the support for discontinuities.

The MUST increase by 1 makes it clear to any distribution participant if there is a gap or missing content. This is a nice guarantee of continuity that is achieved cheaply without requiring additional metadata. The current language in this PR says " _Within a continuous publishing session, each subsequent Group ID MUST increase by 1._", so it is not contradictory with the gap signaling across a restart.

wilaw (approved):

kixelated:
> " _Within a continuous publishing session, each subsequent Group ID MUST increase by 1._", so it is not contradictory with the gap signaling across a restart.

I understand the intent, however:

The meaning of a "continuous publishing session" depends on perspective. What about software that muxes primary/backup input tracks into one output track. Would it be illegal to have a gap in the output track? 

And obviously no subscriber should throw an error if the group ID is not +=1. Group ID gaps can and will happen under normal scenarios due to congestion. I don't like the MUST because it encourages player incorrectly enforce +=1.

wilaw:
What if we said "each subsequent Group ID SHOULD increase by 1. Any intentional gaps MUST be signaled using the MOQT Prior Group ID Gap Extension header." ?

suhasHere:
> The meaning of a "continuous publishing session" depends on perspective. What about software that muxes primary/backup input tracks into one output track. Would it be illegal to have a gap in the output track?

It's the property of output track that we are talking about here. The same rules apply for that track too , saying that when the track is active the group numbers increase by one and when there is a disconnect it does need to start a different point from past.

I feel the current text address this effectively

kixelated:
> What if we said "each subsequent Group ID SHOULD increase by 1. Any intentional gaps MUST be signaled using the MOQT Prior Group ID Gap Extension header." ?

I would have SHOULD for both, but that works for me.

draft-ietf-moq-msf.md
```diff
@@ -1573,19 +1573,16 @@ of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
...
-future, such as after a loss of connectivity or an encoder restart, will not result
-in smaller or duplicate Group IDs for the same track name. Note that this method
-does not prevent duplication if more than 1000 Groups are published per second.
-
-Each subsequent Group ID MUST increase by 1.
-
-If a publisher is able to maintain state across a republish, it MUST signal the gap
-in Group IDs using the MOQT Prior Group ID Gap Extension header.
+Group IDs for a track MUST be unique and MUST increase monotonically. Within a
+continuous publishing session, each subsequent Group ID MUST increase by 1.
```
```suggestion
continuous publishing session, each subsequent Group ID SHOULD increase by 1.
```

Issue #145, opened 2026-03-01
Ordering of delta updates


kixelated:
I mentioned this in another issue, going to split it out:
> The Add, Delete and Clone operations are applied sequentially in the order they are declared in the document. Each operation in the sequence is applied to the target document; the resulting document becomes the target of the next operation. Evaluation continues until all operations are successfully applied.

The operations are keys within a JSON object, so ordering is undefined. Some languages might preserve the ordering, but others will discard it. If you need ordering in JSON, it needs to be an array.

wilaw:
Fixed by https://github.com/moq-wg/msf/pull/158

PR #158, opened 2026-05-11
Replace delta update fields with ordered operations array

suhasHere:
Addresses #145 to enforce ordering of operations. 
If we think we need a diff solution, we can add operation id and do it in the increasing order too ?


wilaw (commented):
Can we simplify this? Currently we have a flag (deltaUpdate) that tells if this catalog represents an update and then another property whose value is an array to carry the operations, that is only present if the flag is TRUE. 

To simplify, let the value of deltaUpdate be the Array of operations. If deltaUpdate is not present, then the catalog is independent. If the property is present, then its value holds an ordered array of operations.

suhasHere:
@wilaw  latest commit should address you suggestion

wilaw (approved):
This is compact and clear.

PR #159, opened 2026-05-11
Add catalog compression support via Track/Object Properties

### LINKED ISSUE #144: Compression for the catalog
We should support at least zlib for catalog entries.  They have a lot of built-in redundancy, especially given we do things like use base64 for init segments.

suhasHere:
@wilaw @vasilvv thoughts on this ?

Fixes #144

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -256,7 +256,8 @@ encrypted. The Catalog provides the names and namespaces of the tracks being
 produced, along with the relationship between tracks, properties of the tracks
 that consumers may use for selection and any relevant initialization data.
 
-The catalog track MUST have a case-sensitive Track Name of "catalog".
+The catalog track MUST have a case-sensitive Track Name beginning with "catalog",
+optionally followed by a compression suffix as defined in {{catalog-compression}}.
 
 A catalog object MAY be independent of other catalog objects or it MAY represent
 a delta update of a prior catalog object. The first catalog object published
@@ -272,6 +273,30 @@ Each catalog update MUST be mapped to an MOQT Object.
 Subscribers accessing the catalog MUST use SUBSCRIBE with a Joining FETCH
 (offset = 0) in order to obtain the latest complete catalog along with all subsequent
 catalog objects, including delta updates, that follow.
+
+## Catalog Compression {#catalog-compression}
+
+Catalogs can contain significant redundancy, particularly when initialization
+data is included. To reduce payload size, the catalog MAY be compressed.
+
+Compression is signaled via the catalog track name suffix:
+
+| Track Name   | Compression | Reference    |
+|:=============|:============|:=============|
+| catalog      | None        | -            |
+| catalog.gz   | GZIP        | {{GZIP}}     |
+
+Table X: Catalog Compression Suffixes
+
+Publishers MAY advertise multiple catalog track variants with different
+compression algorithms. Subscribers MUST subscribe to a catalog variant
+using a compression algorithm they support.
+
+All MSF implementations MUST support both "catalog" (uncompressed) and
+"catalog.gz" (GZIP compressed).
+
+Future compression algorithms can be added to the MSF Compression Suffix
+Registry ({{iana-compression-registry}}).
 
 A catalog is a JSON {{JSON}} document, comprised of a series of mandatory and
 optional fields. At a minimum, a catalog MUST provide all mandatory fields. A
@@ -2386,6 +2411,20 @@ The initial contents of this registry are:
 | urn:scte:scte35:2013:xml       | SCTE-35 XML representation         | {{SCTE35-MSF}}   |
 | urn:msf:timedtext:webvtt        | WebVTT timed text cues                | {{WebVTT-MSF}}   |
 | urn:msf:timedtext:imsc1         | IMSC1 timed text  cues                 | {{IMSC1-MSF}}    |
+
+## MSF Compression Suffix Registry {#iana-compression-registry}
+
+IANA is requested to create a new "MSF Compression Suffix Registry" with
+the following initial entries:
+
+| Suffix | Compression Algorithm | Reference | Status   |
+|:=======|:======================|:==========|:=========|
+| (none) | None                  | RFC XXXX  | MTI      |
+| .gz    | GZIP                  | {{GZIP}}  | MTI      |
+| .zst   | Zstandard             | RFC 8878  | Optional |
+| .br    | Brotli                | RFC 7932  | Optional |
+
+New entries require Specification Required policy per RFC 8126.
 
 --- back
 
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -281,7 +281,7 @@ data is included. To reduce payload size, the catalog MAY be compressed.
 Compression is signaled via the MSF_CATALOG_COMPRESSION MOQT Track Property
 (see {{MoQTransport}} Section 14.4). Track properties are carried in MOQT
 control messages, allowing endpoints to learn the compression algorithm
-before receiving any catalog object payload. The property value is a varint
+before receiving any catalog object. The property value is a varint
 indicating the compression algorithm:
 
 | Value | Compression Algorithm | Reference |
@@ -292,22 +292,11 @@ indicating the compression algorithm:
 Table X: Catalog Compression Values
 
 Publishers that compress the catalog MUST include the MSF_CATALOG_COMPRESSION
-track property in control messages as follows:
+track property in the PUBLISH message (publisher-initiated flow) or
+SUBSCRIBE_OK (subscriber-initiated flow).
 
-* When the publisher initiates the flow using a PUBLISH message to announce
-  the catalog track, the property MUST be included in the PUBLISH message.
-* When responding to a SUBSCRIBE message from a subscriber, the property
-  MUST be included in the SUBSCRIBE_OK response.
-
-Publishers MUST NOT omit this property when the catalog is compressed.
-
-Subscribers MUST check for the presence of the MSF_CATALOG_COMPRESSION track
-property before processing the catalog payload:
-
-* When subscribing to the catalog track, subscribers MUST examine the
-  SUBSCRIBE_OK message for this property.
-* When receiving a catalog track via a publisher-initiated flow, subscribers
-  MUST examine the incoming announcement for this property.
+Subscribers MUST check for this property in the corresponding message before
+processing the catalog payload.
 
 If the property is absent, the subscriber MUST treat the catalog as
 uncompressed. If the property is present with a value the subscriber does
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -283,40 +283,6 @@ A parser MUST ignore fields it does not understand.
 
 ## Root Catalog Fields
 
-
-## Catalog Compression {#catalog-compression}
-
-Catalogs can contain significant redundancy, particularly when initialization
-data is included. To reduce payload size, the catalog MAY be compressed.
-
-Compression is signaled via the MSF_CATALOG_COMPRESSION MOQT Track Property
-(see {{MoQTransport}} Section 14.4). Track properties are carried in MOQT
-control messages, allowing endpoints to learn the compression algorithm
-before receiving any catalog objects. The property value is a varint
-indicating the compression algorithm:
-
-| Value | Compression Algorithm | Reference |
-|:======|:======================|:==========|
-| 0     | None (uncompressed)   | RFC XXXX  |
-| 1     | GZIP                  | {{GZIP}}  |
-
-Table X: Catalog Compression Values
-
-Publishers that compress the catalog MUST include the MSF_CATALOG_COMPRESSION
-track property in the PUBLISH message (publisher-initiated flow) or
-SUBSCRIBE_OK (subscriber-initiated flow).
-
-Subscribers MUST check for this property in the corresponding message before
-processing the catalog payload.
-
-If the property is absent, the subscriber MUST treat the catalog as
-uncompressed. If the property is present with a value the subscriber does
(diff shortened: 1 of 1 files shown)


vasilvv:
On one hand, this makes sense.  On the other hand, this makes writing a player more challenging (you have to figure out if you subscribe to `catalog` or `catalog.gz`).

What about adding a varint in front of object 0 specifying the compression algorithm used?  We could use 0x7b if you want to be backwards-compatible.

wilaw:
I'm not a fan of the file extension approach. It reminds me of Windows 95. And as Victor pointed out, we lose the well-known name of "catalog". 

In MOQT, we have track properties as a standard way of adding metadata. Why don't we define a custom property which carries a varint which specifies the type of compression? This would be read by the player before processing the payload. MSF can define this property. if the property is absent then its uncompressed for backwards compatibility. By setting up an IANA table to allow for unconstrained compression types, we put a lot of complexity for player developers who must constantly check the table to see if new entries exist and then update their players. I'd rather we pick one or two compression types and hard-code them.  I'd be fine with just GZIP as its broadly supported.  If we want real compression, we need to get away from JSON in the first place.

vasilvv:
I like the track property idea.  We should make it mandatory, though, or we will encounter a risk that some players will assume it's always uncompressed.

suhasHere:
Agree .I had special preamble as one of the notes, but I am liking track property better too. Let me work on that direction 

Thanks @wilaw @vasilvv

kixelated:
I like using an extension for namespaces. ex. `demo/bbb.msf` is a hint that I should subscribe to `catalog`. 

Otherwise, a generic CDN dashboard can't tell the difference between media and non-media broadcasts, to do stuff like embed a player. Blindly trying to subscribe to all possible catalog tracks is a bad option.

A track property that is the equivalent of a mime/type would work too.

kixelated:
As for compression, I'm fine with it but IDK maybe it should be at the moq-transport layer. Many different track encodings will want some form of compression and we should do it once. Same concept with HTTP.

suhasHere:
@wilaw @vasilvv does the new changes look about right ?

wilaw:
> I'm fine with it but IDK maybe it should be at the moq-transport layer. Many different track encodings will want some form of compression and we should do it once

If MOQT adds a transport layer payload compression identifier, then it can coexist with the track property defined here. Until then, this lets us proceed with catalog compression without waiting for MOQT.

wilaw:
@suhasHere  - I was just about to merge, however shouldn't we allow this compression property to also apply to Media Timeline and Event Timeline tracks? Can you redefine it as a generic track property for MSF (perhaps a new section for Track Properties, of which this is the first) and then reference it for use in catalogs, media timelines and event timelines? Event timelines, with a lot of repeated JSON field names, would compress quite nicely.

suhasHere:
@wilaw you are so right. let me fix that

kixelated:
Can we run some tests/benchmarks first? I'm not even sure if compressing the timeline tracks would save bytes? Who needs this right now?

kixelated:
ESPECIALLY if this is mandatory to implement because we're not doing two tracks

suhasHere:
@wilaw  the changes now should address your recent suggestions.

Also I deployed a simple web-app to test the benchmarking with gzip. Please give it a try at https://quicr.github.io/moq-web/apps/msf-compression-benchmark/

I larger the catalog, timelines , the better the compression efficiency as expected

Screenshot 2026-05-13 at 10 42 00 PM

kixelated:
> Also I deployed a simple web-app to test the benchmarking with gzip. Please give it a try at https://quicr.github.io/moq-web/apps/msf-compression-benchmark/
> 
> I larger the catalog, timelines , the better the compression efficiency as expected

Are you making an object for each timeline update? Of course the entire timeline will compress well, but I'm not sure that's true of the deltas?

wilaw:
@Suhas - love the benchmark tool. It illustrates that long running media and event timelines are practical when compression is applied. 

One final comment. We don't need to register a Track Property in the MOQT IANA table. That table reserves a range of application-specific properties which don't need to be registered, as they are defined by the application. 

> 0x78 to 0x7F: Reserved for application-specific use (1-byte encoding, no registration permitted)

I think we should reserve 0x78 (the first even property available) for MSF_COMPRESSION and define that in our spec. That way people can start using it immediately. 

I'll go ahead and suggest revisions to the sections which we can remove.

wilaw (approved):

draft-ietf-moq-msf.md
```diff
@@ -2355,6 +2363,42 @@ The specific error codes and retry semantics are defined by the authorization
 scheme specifications. See {{PrivacyPassAuth}} for Privacy Pass error handling
 and {{C4M}} for CAT error handling.
 
+# MSF Track Properties {#track-properties}
+
+MSF defines MOQT Track Properties (see {{MoQTransport}} Section 14.4) to signal
```
```suggestion
MSF defines MOQT Track Properties (see {{MoQTransport}} Section 15.8) to signal
```

draft-ietf-moq-msf.md
```diff
@@ -2355,6 +2363,42 @@ The specific error codes and retry semantics are defined by the authorization
...
+(subscriber-initiated flow).
+
+Subscribers MUST check for this property in the corresponding message before
+processing the track payload.
+
+If the property is absent, the subscriber MUST treat the payload as uncompressed.
+If the property is present with a value the subscriber does not support, the
+subscriber MUST NOT attempt to process the payload and SHOULD unsubscribe from
+the track.
+
```
```suggestion
the track.

The MSF_COMPRESSION property ID is 0x78.

```

draft-ietf-moq-msf.md
```diff
@@ -2387,6 +2431,30 @@ The initial contents of this registry are:
...
+| Property Name   | Property ID | Value Type | Reference |
+|:================|:============|:===========|:==========|
+| MSF_COMPRESSION | TBD         | varint     | RFC XXXX  |
+
+The MSF_COMPRESSION property indicates the compression algorithm applied to
+JSON-based MSF track payloads. See {{compression-property}} for the full
+specification.
+
+This document also requests IANA to create a new "MSF Compression Algorithms"
```
```suggestion
This document requests IANA to create a new "MSF Compression Algorithms"
```

wilaw (commented):
I have suggested the changes which enable us to define the MSF_COMPRESSION property locally and not have to register it with IANA.

vasilvv:
I am a bit skeptical about the media timeline too.  But I already said before that I think the current format is very suboptimal.

I imagine event timeline would depend on the nature of events.  Some compress really well, and for some it does not help at all.

> Also I deployed a simple web-app to test the benchmarking with gzip. Please give it a try at https://quicr.github.io/moq-web/apps/msf-compression-benchmark/

I think this overestimates the gain from the timeline tracks (we'd want to publish individual entries to reduce latency, so we can't compress them as a single blob), and underestimates the gain from the catalog (it assumes init segment is random bytes, but in practice those are not).

suhasHere:
@vasilvv i agree with your conclusion. The tool is not perfect . It's purely random and but it show gains with compression.
Also on initData , the tool allows you to not have it in the tracks.

On large vs small sizes of raw data , from my experience with dealing with SDP, everything started small, until it got big. Also we need to supprt flows where someone wants to download timelines and catalog from a long history. 

I would want to better understand if we think if there are any blockers for not to proceed forward in this direction?

vasilvv:
My personal opinion is that we should land compression for catalog tracks for now, and open a separate issue discussing whether it makes sense for timelines.

kixelated:
Yeah I was imagining that timeline delta/update would be maybe 10-30 bytes. They'll often just be a mapping from group -> timestamp. The GZIP header is probably larger at that point.

wilaw:
> My personal opinion is that we should land compression for catalog tracks for now, and open a separate issue discussing whether it makes sense for timelines.

Compression for timelines makes sense if I have a long timeline for say a 6 hr cricket match, with a group every 2s. Thats 10,800 JSON records. Compression certainly helps in this case. The player needs to load that if it wants to construct a scrub-bar UI or if it joins the stream late. Compression is optional. If it helps your particular scenario, then use it, otherwise don't. 

If you are using media timelines with a template, don't use it.
If you are sending delta updates, don't use it. 
If your event timeline is only 5 records long, don't use it. 

The point about delta updates makes me think however - should this compression property be a Track property, or an Object property? Consider the case of the long cricket game, which is using delta updates every 2s, but every 5 min it repeats the entire 6hr history. If I run a small delta payload [`8008, [4,0], 1759924166389]` through Suhas's tool, it estimates a -71% compression efficiency i.e payload increases under compression. In that case we would want compression on the group updates but not the individual delta updates. I'm thinking the compression should be specified per Object ? It would be an easy change to switch it to an Object property.

suhasHere:
> The point about delta updates makes me think however - should this compression property be a Track property, or an Object property? Consider the case of the long cricket game, which is using delta updates every 2s, but every 5 min it repeats the entire 6hr history. If I run a small delta payload [`8008, [4,0], 1759924166389]` through Suhas's tool, it estimates a -71% compression efficiency i.e payload increases under compression. In that case we would want compression on the group updates but not the individual delta updates. I'm thinking the compression should be specified per Object ? It would be an easy change to switch it to an Object property.

I am with @wilaw  on landing it and an implementation MAY NOT compress.
OTOH, keeping it not do just for mediatimeline and eventtimeline, but allowing it for catalog is not going to help much.  Since if a implementation supports gzip for catalog, it has the code to do for timelines too. Publishers can choose not compress mediatimeline/eventtimeline tracks.

On the object vs track property is 
- Adding track property still makes sense
- For delta updates use-case which is a great find, may be specifying it under deltaUpdates section on how a delta heavy can leverage compression using setting it on first object of group is useful. However, i do feel, it can be its own PR ?

wilaw:
> may be specifying it under deltaUpdates section on how a delta heavy can leverage compression using setting it on first object of group is useful.

But if its set as a track property then its sent in the SUBSCRIBE_OK or PUBLISH_OK message and it applies to **all** future objects in the track, whether they need compression or not. Thats why defining it as an Object property would allow you to set it on only the first Object in group, as you suggested.

suhasHere:
> But if its set as a track property then its sent in the SUBSCRIBE_OK or PUBLISH_OK message and it applies to **all** future objects in the track, whether they need compression or not. Thats why defining it as an Object property would allow you to set it on only the first Object in group, as you suggested.

I think its ok to have property repeated in Object too. We need to specify the rules for it in MSF spec saying it overrides the track level setting.

suhasHere:
@wilaw how about this proposal
1. We define track property 
2. We define object property
Both are in their numbering space, we can use the same property type.

Along with the current text, we will also add text around Object Property and how its get used - can only be set on first object a group and receivers use the compression per that property, if available

kixelated:
> > My personal opinion is that we should land compression for catalog tracks for now, and open a separate issue discussing whether it makes sense for timelines.
> 
> Compression for timelines makes sense if I have a long timeline for say a 6 hr cricket match, with a group every 2s. Thats 10,800 JSON records. Compression certainly helps in this case. The player needs to load that if it wants to construct a scrub-bar UI or if it joins the stream late. Compression is optional. If it helps your particular scenario, then use it, otherwise don't.

Why not have a single group for the entire timeline, with a delta object every 2s? What's the purpose of a group every 2s?

wilaw:
> Compression for timelines makes sense if I have a long timeline for say a 6 hr cricket match, with a group every 2s.
> 
> Why not have a single group for the entire timeline, with a delta object every 2s? What's the purpose of a group every 2s?

You're mixing the groups of the timeline with the 'groups' of the video track. The video track has groups every 2s because it has a keyframe every 2 seconds. That dictates the length of the media timeline track payload, which adds a record for each group.

In thinking about this more, we only need compression when a large payload is sent, whether by group or object. This is the case with a non-delta catalog. Delta catalog updates sent by a relay which each be small and compression would be counter-productive. Same for delta updates for a timeline track. Which leads me back to the thought that compression should be applied per object. The relay has a simple rule - if payload > X then compress, otherwise don't. 

IN fact, I can think of only one case in which compression should be applied at the track level - when it never does delta updates and every payload is large. In that case the track property is saving us from one byte with every large payload, which does not seem worth the complexity of defining two compression headers. I think an object compression header gets us what we need in the simplest way.

wilaw:
> how about this proposal
> 
> We define track property
> We define object property
> Both are in their numbering space, we can use the same property type.
> Along with the current text, we will also add text around Object Property and how its get used - can only be set on first object a group and receivers use the compression per that property, if available

@suhasHere  - lets go with the dual Object and track properties definition. I don't see a use-case where you use both. You use the Track property if every single object is compressed, otherwise you signal individual compression at the object level if you have a mixture of compressed and uncompressed objects in your track.  Only setting it on the first object in a group subverts the catalog use-case, where we want the first catalog in a group to be compressed because its the complete catalog and then the delta updates that follow to be uncompressed. So lets keep it simple - all objects are compressed, or else individual Objects signal compression. If you make these changes, we can merge.

tobbee:
I think this is fine and simpler than having an Accept Encoding" subscribe parameter that I suggested in a comment in #144.

suhasHere:
@wilaw latest commit has the new changes.

draft-ietf-moq-msf.md
```diff
@@ -2355,6 +2367,78 @@ The specific error codes and retry semantics are defined by the authorization
...
+# MSF Properties {#track-properties}
+
+MSF defines MOQT Track Properties and Object Properties (see {{MoQTransport}})
+to signal metadata about MSF tracks and objects. These properties are carried in
+MOQT control messages and object headers, allowing endpoints to learn track and
+object characteristics before processing payload data.
+
+## Compression Signaling {#compression-signaling}
+
+MSF provides two mutually exclusive mechanisms to signal compression of JSON-based
```
wilaw: This compression is useful for non-JSON payloads which may be added in the future. No need to limit it now. 

```suggestion
MSF provides two mutually exclusive mechanisms to signal compression of
```

wilaw (approved):

PR #165, opened 2026-05-25
Update bitrate and related properties in draft

### LINKED ISSUE #164: Require sample rate and channels
These fields should be required for audio tracks. If they're optional, I have to parse the init segment (gross) just to figure out if I should subscribe to the given track.

And yeah I already filed a few issues, but we should have more required fields in MSF. It's reaally annoying that everything is optional.

wilaw:
1. Adds Maximum GOP Duration track property
2. Adds Maximum Group Duration track property
3. Adds Average Bitrate track property
4. Makes sample rate and channels properties required for audio tracks. 
5. Makes codec, width, height required for video tracks. 
6. Redefines "bitrate" specify the maximum bitrate

Fixes #164

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -384,7 +384,10 @@ Table 2 lists the fields defined within each track object.
 | Mime type               | mimeType               | {{mimetype}}              |
 | Framerate               | framerate              | {{framerate}}             |
 | Timescale               | timescale              | {{timescale}}             |
-| Bitrate                 | bitrate                | {{bitrate}}               |
+| Maximum bitrate         | bitrate                | {{bitrate}}               |
+| Average bitrate         | avgBitrate             | {{averagebitrate}}        |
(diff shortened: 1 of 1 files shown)


vasilvv:
> Makes codec, width, height required for video tracks.

This seems odd, I thought we discussed before that we wanted to make width/height optional as the actual width and height might vary over the duration of the track?

kixelated:
I agree the current width/height can change (unless using initData).

I think you should know the max width/height, at the very least for determining which track best fits the viewport. Technically the codec compatibility (part of the mime) does advertise the encoder's maximum, but it's usually inflated.

wilaw:
Commit https://github.com/moq-wg/msf/pull/165/changes/824bba8cf063d1dc8d12597ad59a67143c8d42de resets width and height to optional. it also defines them to be the max likely to be encountered over the life of the track in the case that they do vary.

suhasHere (approved):

Issue #153, opened 2026-04-10
`initTrack` does not work


vasilvv:
Right now, if the publisher emits a new init group, and a new media group that requires that init group (let's say both are group 500), there is no way for the receiver to know if a new init group has been generated.  If the receiver gets a media object for group 500, it can incorrectly apply the old init segment, since there is no guarantee a new one arrives first (as those are on different tracks).

I can see three potential ways to fix it:
1. Add a field/extension for MOQT objects that indicates the version of the init segment required, and make client sync those (this has to work for every container).
2. Remove initTrack.
3. Remove the ability to update initTrack (only allow group 0 object 0).

I'm personally in favor of the approach 2, since I don't think the extra complexity is worth it.

wilaw:
This is a fair comment. We should fix this now before the next draft. Rather than remove, I propose we adjust the design. Here is another proposal for achieving init declaration compactness but also allowing for mi-stream changes. 

One of the stated benefits of init tracks is to reduce catalog bloat, as we end of repeating the relatively large init segments for each track, even when identical. In CMSF, we introduced [contentProtections ](https://moq-wg.github.io/cmsf/draft-ietf-moq-cmsf.html#name-content-protections)which is an array of contentProtections fields, each with an ID which can then be referenced in each track definition. Since contentProtection elements are often re-used by tracks, this referencing is efficient.

We could do the same for init segments. Declare an inits array in the root and list a bunch of inits ,each with a reference ID. The track definition then just links to the init ID. This design also has the advantage that we can declare inits which are not used now but may be used later. 

```
inits [
 "1": "kjg34hg4....84h4nbrb",
 "2": "mn34mn...vc287dhgds",
 "3": "lkj459856...bvsndbfv73=",
]
```

For switching the init mid stream, we need a mechanism to give the client a warning. For this we can define a derivative of the event timeline track, used location-based indexing. If the publisher wants to switch to init #3 at group 34, then a few groups before that ( at least longer than the players forward buffer) it would update the catalog to include a reference to init #3 and also publish an update to its init timeline track. That timeline track would define the precise group at which the replacement must occur. 

```
[
    {
        "l": [0,0],
        "data": "1"
    },
    {
        "l": [34,0],
        "data": "3"
    }
]
```

vasilvv:
I don't think this really solves the race condition, since the new track can still arrive out-of-order.  As far as I can tell, the only solution that works is adding in-band signaling to the actual media payload track.

wilaw:
[...]

Issue #136, opened 2026-02-27
No mechanism to delta update a track


kixelated:
There's no way to update a field in a track? I currently do that when the width/height of the media changes, such as when a shared window is resized.

In theory I could remove/add, but current draft doesn't specify the ordering of remove/add. But I don't want to reserialize the entire track information anyway (especially the init data) just to update one field.

In theory you could clone the existing track... but that would temporarily cause two tracks with the same name. It's not clear if this is illegal. If it's not, we would need a remove in the same delta update, but the ordering would have to be `clone/remove/add`.

wilaw:
Since the OP, we have
1. Defined width/height to be the max that might be transmitted by the track. This may remove the need to update width/height when a shared window is resized. 
2. The cloning operation is now an ordered array, so you can clone -> remove -> add in a logically consistent sequence. This allows you to cheaply change a subset of properties.

PR #166, opened 2026-05-25
Enhance root catalog with Initialization Data List

wilaw:
Added Initialization Data List and updated related fields.

The purpose of this change is to make the catalog more human readable by grouping all init data and placing it at the end of the JSON document. 

Partial fix for #153

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -376,7 +376,7 @@ object has the following fields:
 |:==============|:===============================================|
 | inline        |  Base64 {{BASE64}} encoded initialization data |
 
-* data: a string holding the init payload as defined by the type. 
+* data: a string holding the init payload as defined by the type.
 
 The Initialization Data List, if present, MUST be located after the tracks array in
 the root of the JSON catalog. The purpose of this is to improve the human readability
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -533,7 +533,7 @@ a set video tracks of the same content offered in alternate bitrates.
 ### Initialization reference {#initref}
 Required: Optional    JSON Type: Number    Location: Track Object
 
-A number pointing at the zero-based index of an entry in the Intialization
+A number pointing at the zero-based index of an entry in the Initialization
 Data List {{initdatalist}}.
 
 ### Dependencies {#dependencies}
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -285,14 +285,15 @@ A parser MUST ignore fields it does not understand.
 
 Table 1 lists the fields defined at the root of the catalog JSON object.
 
-| Field                   |  Name                  |           Definition      |
-|:========================|:=======================|:==========================|
-| MSF version             | version                | {{msfversion}}            |
-| Generated at            | generatedAt            | {{generatedat}}           |
-| Is Complete             | isComplete             | {{iscomplete}}            |
-| Tracks                  | tracks                 | {{tracks}}                |
-| Publish tracks          | publishTracks          | {{publishtracks}}         |
-| Delta update            | deltaUpdate            | {{deltaupdate}}           |
+| Field                     |  Name                  |           Definition      |
+|:==========================|:=======================|:==========================|
+| MSF version               | version                | {{msfversion}}            |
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -518,10 +530,11 @@ sequences. A subscriber typically subscribes to one track from a set of
 tracks specifying the same alternate group number. A common example would be
 a set video tracks of the same content offered in alternate bitrates.
 
-### Initialization data {#initdata}
-Required: Optional    JSON Type: String    Location: Track Object
+### Initialization reference {#initref}
+Required: Optional    JSON Type: Number    Location: Track Object
 
-A string holding Base64 {{BASE64}} encoded initialization data for the track.
+A number pointing at the zero-based index of an entry in the Intialization
```
```suggestion
A number pointing at the zero-based index of an entry in the Initialization
```

tobbee:
I think this is a good start to both increase readbility and let multiple tracks use the same initData.

However, it is quite hard to extend and a bit fragile by using a list index instead of named entries.

Here are a couple of possible improvements:

**1. Make each list entry a typed object, not a bare string.**

A bare base64 string has no room for a future "the bytes live somewhere else" option (e.g. a separate MoQ track, an HTTP URL, a content-addressed identifier). Once initDataList is defined as Array<string>,  the only way to add an alternative resolution mechanism is to either reuse the string with a prefix convention or to introduce a parallel field — both worse than just having a discriminator from day one.

 Suggested shape (with a future possible extension where the initData is in another track).

```json
  "initDataList": [
     { "id": "v1", "type": "inline", "data": "AAAA…base64…" },
     { "id": "v2", "type": "moqRef",
      "track": ["example.com","initdata"], "group": 1, "object": 0 }
  ]
```

with the per-track field referring to id instead of an integer index: `initRef": "v1"`

Unknown type values (and corresponding tracks) are skipped by older parsers; tracks referring to them are simply unusable for that receiver. This is the same forward-extensibility pattern CMSF contentProtections already uses.


**2. Named refs (string id) rather than numeric indices.**

Even without the typed-object change, replacing `"initRef": 0` with `initRef: "v1"` makes catalog edits safer: appending, removing, or reordering entries doesn't shift every other track's initRef value.
It also reads better in JSON without needing to try to find where the n'th base64 blob starts in a list, one
can instead search for "v1".
 
Both points map cleanly onto the contentProtections precedent (root-level array of named entries; per-track …`RefIDs: ["widevine"]`), so the catalog editor model stays consistent across DRM and init data.

*3. Future extension: time-varying init (not to be included now)*

This PR is intentionally scoped to the static case, which is fine. But there is a real live-encoding case where init bytes change mid-stream (parameter-set refresh, codec-config update) as brought up in #153.

I think it certainly needs more discussion, but with the typed/named shape above, a follow-up PR can add a per-track schedule form using the same references.

```json
  "initSchedule": [
    { "fromGroup": 0,     "ref": "v1" },
    {"fromGroup": 5400,  "ref": "v2" }
  ]
```

This could possibly also be used to change other parameters as time progresses as was brought up in #136.
In any case, this is another future  PR, if any.

wilaw:
@tobbee  - see the latest commits. I have revised the indexing to follow the id | type  | data structure you propose. I agree that this is more extensible. I have only allowed the single type of "inline" for now, as we previously had init tracks and then rolled them back for various reasons. So we can add them in later if we have implementation demand. For now, the structure is extensible, but still simple. 

>   "initDataList": [
>      { "id": "1", "type": "inline", "data": "AAAA…base64…" },
>      { "id": "2", "type": "inline", "data": "BBBB…base64…" },
>   ]

tobbee:
@wilaw That's perfect. I didn't wanted to introduce the init tracks now, but just make the format future extensible.

tobbee (approved):
Looks like the right level right now.
Only one type of initData available, but opens up for future extensibilitiy.

gwendalsimon (approved):
I agree it is better to have initdata in a specific location in the catalog. I like the format as it is.
Next step would be to update the examples to reflect the change.

wilaw:
@gwendalsimon  - the examples in the MSF spec are all LOC packaging, in which init data is carried via Object Properties., so unaffected by this change. The examples we need to update are in the CMSF spec. To do that, we must first publish a new draft of MSF, update CMSF to point at that and then update the CMSF examples.

PR #167, opened 2026-05-25
Introduce target buffer property in track object

### LINKED ISSUE #150: Wall clock is problematic
> [5.1.16. ](https://www.ietf.org/archive/id/draft-ietf-moq-msf-00.html#section-5.1.16)[Target latency](https://www.ietf.org/archive/id/draft-ietf-> moq-msf-00.html#name-target-latency)
> Location: T Required: Optional JSON Type: Number
>
>The target latency in milliseconds. Target latency is defined as the offset in wallclock time between when content was encoded and when it is > displayed to the end user. For example, if a frame of video is encoded at 10:08:32.638 UTC and the target latency is 5000, then that frame > should be rendered to the end-user at 10:08:37.638 UTC. This field MUST NOT be included if isLive is FALSE. All tracks belonging to the same > render group MUST have identical target latencies. All tracks belonging to the same alternate group MUST have identical target latencies. If this field is absent from the track definition, then the player MAY choose the latency with which it renders the content.

This will fail without clock synchronization between the publisher and every viewer. It's very common to have clients with stale/disabled NTP synchronization.

 And even then, this target latency is not what you want. Viewers further from the origin will have smaller jitter buffers, hurting the user experience. It's fine to have somebody in Europe 100ms behind somebody in America (if the origin is there); don't try to counter the speed of light.

Target latency should dictate the size of the jitter buffer. Any form of wall clock synchronization between viewers needs something more sophisticated IMO.

wilaw:
Add target buffer property to track definition with requirements.

Fixes #150

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -372,6 +372,7 @@ Table 2 lists the fields defined within each track object.
 | Event timeline type     | eventType              | {{eventtype}}             |
 | Is Live                 | isLive                 | {{islive}}                |
 | Target latency          | targetLatency          | {{targetlatency}}         |
+| Target buffer           | targetBuffer           | {{targetbuffer}}          |
 | Track role              | role                   | {{trackrole}}             |
 | Track label             | label                  | {{tracklabel}}            |
 | Render group            | renderGroup            | {{rendergroup}}           |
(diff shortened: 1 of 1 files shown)


kixelated:
This is good for Sye-like stuff. A minimumBuffer and maybe maximumBuffer would also be worth-while. 

ex. minumumBuffer would need to be at least the CMAF fragment size. maximumBuffer could be the amount of time a segment should be cached, IDK.

wilaw:
Behold fc4a37de3e8ef37ffc64bfe05f3c43faa47a4bff, we now have target, min and max buffers.

tobbee:
OK. Seems like a good start. The syntax with just numbers rather than a map with keys, is a bit hard to read in my view, but I can live with it.

kixelated:
> OK. Seems like a good start. The syntax with just numbers rather than a map with keys, is a bit hard to read in my view, but I can live with it.

Yeah I would do a JSON object too

wilaw:
Agree that a JSON object is clearer, and also more extensible if we want to add different buffer thresholds in the future. Latest commit add this:

`"buffers": {"target":2000, "min": 1500, "max": 5000},`

Please approve so we can merge.

tobbee (approved):
Looks good with the JSON object

PR #168, opened 2026-05-25
Revise catalog object specifications and numbering

### LINKED ISSUE #149: Catalog Mapping to MoQT
> A catalog object MAY be independent of other catalog objects or it MAY represent a delta update of a prior catalog object. The first catalog object published within a new group MUST be independent. A catalog object SHOULD be published only when the availability of tracks changes.
>
> Each catalog update MUST be mapped to an MOQT Object.

If I understand this correctly, `deltaUpdate` MUST not be present for the first object within a group, but otherwise there's no restrictions.

- What about sub-groups?
  - Do deltas only apply in the same sub-group?
  - Do you use Object ID to determine the latest sub-group and (temporarily) ignore the others?
- What if the first object ID is not 0?
- Can independent catalog objects follow delta catalog objects? (on the same stream/sub-group)

wilaw:
Clarify catalog object requirements and update Object ID rules.

Fixes #149

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -261,23 +261,29 @@ The catalog track MUST have a case-sensitive Track Name of "catalog".
 A catalog object MAY be independent of other catalog objects or it MAY represent
 a delta update of a prior catalog object. The first catalog object published
 within a new group MUST be independent and MUST provide a complete catalog that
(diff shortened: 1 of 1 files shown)


tobbee (approved):
Looks fine to me.

gwendalsimon (approved):
it is good

PR #169, opened 2026-05-26
Update MOQT mapping details in media transmission section

### LINKED ISSUE #148: Media Mapping to MoQT

wilaw:
Clarified mapping of MOQT Groups and Objects to Streams.

Fixes #148

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1712,9 +1712,9 @@ In this example:
   as defined in the respective specifications.
 
 # Media transmission
-The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
-of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
-MOQT Stream.
+The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. This mapping is
+defined by the packaging {{mediapackaging}} type of each track. Irrespective of the
+packaging type in place, each MOQT Object MUST be placed on MOQT sub-group 0.
 
 ## Group numbering
 Group IDs for a track MUST be unique and MUST increase monotonically. Within a
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -167,8 +167,8 @@ MSF delivers LOC {{LOC}} packaged media bitstreams.
 This specification references Low Overhead Container (LOC) {{LOC}} to define how
 audio and video content is packaged. With this packaging mode, each
 EncodedAudioChunk or EncodedVideoChunk sample is placed in a separate MOQT
-Object. Samples that belong to the same Group of Pictures (GOP) MUST be placed
-within the same MOQT Group.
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -1712,9 +1712,9 @@ In this example:
   as defined in the respective specifications.
 
 # Media transmission
-The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. Irrespective
-of the {{mediapackaging}} in place, each MOQT Object MUST be mapped to a new
-MOQT Stream.
+The MOQT Groups and MOQT Objects need to be mapped to MOQT Streams. This mapping is
+defined by the packaging {{mediapackaging}} type of each track. Irrespective of the
+packaging type in place, each MOQT Object MUST be placed on MOQT sub-group 0.
```
suhasHere, line 1717: this does limit all media use-cases to be having only one group isn't it?

wilaw: I don't understand that comment. The text does not mention Groups at all. All it is doing is constraining usage to use sub-group 0 within each Group. There can be as many Groups as the packaging format sees fit.

gwendalsimon:
I am in favor of merging this PR and discarding #156. #156 signals Subgroup
layout without the dependency information a receiver needs to act on it. The only
use case I see for multiple Subgroups is SVC, and extending MSF to cover it would
be a very good idea. I am not volunteering to do it.

The PR is not right as written: the current lead sentence is confusing and does
not state the point clearly. Suggested text:

> Within each MOQT Group, all Objects MUST be assigned to a single MOQT Subgroup
> whose Subgroup ID is 0, unless an extension or packaging specification defines
> a different Subgroup structure. The mapping of media samples to MOQT Objects
> and Groups is defined by the packaging specification ({{mediapackaging}}); the
> mapping of Subgroups to QUIC streams is defined by [MoQTransport].

The "unless an extension or packaging specification defines a different Subgroup
structure" clause is deliberate: it acknowledges that multiple Subgroups can
exist (for example, SVC layers) and leaves room for a future extension to define
them, without the base draft needing to say more.

suhasHere:
Vienna 
 Makes no sense to prohibit multiple sub-groups

PR #174, opened 2026-05-27
Update media presentation timestamp rounding description

### LINKED ISSUE #108: timescale in timeline track
We could add a timescale ticks as a parameter for media timelines

Or we can say that media_PTS is millisecond precision (floor), and more precise microsecond can be obtained from the FETCHed object.

The second option get consensus.

wilaw:
Clarified timestamp rounding requirement for media presentation.

Fixes #108

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1778,8 +1778,7 @@ The explicit format contains an array of records. Each record consists of
 an array of three required items, whose ordinal position defines their type:
 
 * The first item holds the media presentation timestamp, expressed as a JSON Number.
-  This value MUST match the media presentation timestamp, rounded to the floor of the
-  nearest millisecond, of the first media sample in the referenced Object. Implementers
+  This value MUST match the media presentation timestamp, expressed as the floor in integral milliseconds, of the first media sample in the referenced Object. Implementers
   who require increased time precision can parse the retrieved media object itself.
 * The second item holds the MOQT Location of the entry, defined as a tuple of the MOQT
   Group ID and MOQT Object ID, and expressed as a JSON Array of Numbers, where the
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -1778,8 +1778,9 @@ The explicit format contains an array of records. Each record consists of
 an array of three required items, whose ordinal position defines their type:
 
 * The first item holds the media presentation timestamp, expressed as a JSON Number.
-  This value MUST match the media presentation timestamp, rounded to the nearest
-  millisecond, of the first media sample in the referenced Object
+  This value MUST match the media presentation timestamp, rounded to the floor of the
```
tobbee: I cannot make sense of "rounded to the floor of the **nearest** millisecond". 

I see two roundings: 

1. "nearest" which I interpret as a full millisecond (possibly a little in the future)
2. "floor" that rounds down (not in future). Meaningless since we have already chosen the nearest

I think we should say something like

This value MUST match the media presentation timestamp rounded to integral milliseconds, of ...

Or, "... expressed as the floor in integral milliseconds, of ..." if the floor is important.

One could move the integral part up to the sentence saying that it should be "..., expressed as an integral JSON Number."

tobbee (commented):
I think the rounding wording is weird and lack clarity in that it should be an integral number in milliseconds.

draft-ietf-moq-msf.md
```diff
@@ -1778,8 +1778,9 @@ The explicit format contains an array of records. Each record consists of
...
+  nearest millisecond, of the first media sample in the referenced Object. Implementers
```
```suggestion
  This value MUST match the media presentation timestamp, expressed as the floor in integral milliseconds, of the first media sample in the referenced Object. Implementers
```

wilaw:
@tobbee  - https://github.com/moq-wg/msf/pull/174/changes/5f83998d72eaadbef164ed3493104356ed8b6aea applies your suggestion

tobbee (approved):
I'm OK now.

PR #190, opened 2026-07-15
Add Event Timeline track Property to carry header data

### LINKED ISSUE #180: LondonInterim: Should event timeline track have a header section
Intriduce a track property for eventtimeline header

wilaw:
Added section on Event Timeline header data and properties, detailing requirements for header data in event payloads and the MSF_EVENT_TIMELINE_HEADER Track property.

Fixes #180

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -1985,6 +1985,20 @@ Event timeline tracks are optional. Multiple event timeline tracks can exist ins
 catalog. The type & structure of the data contained within each event timeline track is
 declared in the catalog, to facilitate client selection and parsing.
 
+## Event Timeline header data {#eventtimelineheader}
+Certain event payloads require header data to help initialize or give context to the
+payload data, which consists solely of an array of records. This header data is typically
+invariant over the life of the track.
+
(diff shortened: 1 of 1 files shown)


gwendalsimon:
It is a good contribution, but not easy to understand. An example may help

gwendalsimon:
An example that could close the new section, building on the drone telemetry already in "Event timeline track with MOQT Location indexing".

The catalog declares the track and its type:

```json
{
  "name": "telemetry",
  "namespace": "live.example.com/flight/1",
  "packaging": "eventtimeline",
  "eventType": "com.example.dronetelemetry",
  "depends": ["1080p-video"]
}
```

The publisher sends the invariant context in the `MSF_EVENT_TIMELINE_HEADER` track property, serialized as the event type defines:

```json
{"crs": "WGS84", "fields": ["latitude", "longitude", "altitude"], "altitudeUnit": "meter"}
```

Each payload record then carries values alone:

```json
[
  {"L": [0,0], "data": [47.1812, 8.4592, 488.2]},
  {"L": [1,0], "data": [47.1662, 8.5155, 502.7]}
]
```

Without the header a subscriber cannot tell latitude-first ordering from longitude-first, nor meters from feet. Repeating the three field names in every record would cost more than the values themselves, which is the case for a track property.

PR #210, opened 2026-09-07
Refactor track namespace and name definitions

### LINKED ISSUE #45: Should the catalog format constrain the grammar of track namespaces and track names?
MOQ Transport says:

>    In this specification, both the Track Namespace and the Track Name
>   are not constrained to a specific encoding.  They carry a sequence of
>   bytes and comparison between two Track Namespaces or Track Names is
>   done by exact comparison of the bytes.  Specifications that use MoQ
>   Transport may constrain the information in these fields, for example
>   by restricting them to UTF-8.  Any specification that does needs to
>   specify the canonicalization into the bytes in the Track Namespace or
>   Track Name such that exact comparison works.

Should the catalog format indeed constrain the grammar of these fields, whether just be restricting them to be valid UTF-8 or somehow further?  (Valid URI?  ASCII-Only?). Otherwise we need to deal with arbitrary binary data in the track namespaces and track names, which can be challenging to represent in JSON.

### LINKED ISSUE #209: Catalog: mapping of JSON name / namespace strings to byte sequences is unspecified
The catalog carries `name` and `namespace` as JSON strings and says nothing about how the JSON string becomes the bytes for MOQT. 

  ### Unspecified

  - Whether the Track Name bytes are the UTF-8 encoding of the JSON string value at all. 
  - If UTF-8, whether the bytes are taken verbatim, i.e. neither producer nor parser may apply Unicode normalization. 
  - How to handle `\u` escapes
      - Whether `"caf\u00e9"` and a raw `"café"` denote the same track. 
      - How to handle unpaired surrogate escapes such as `\uDEAD`. [RFC 8259 8.2](https://www.rfc-editor.org/rfc/rfc8259#section-8.2) permits them in JSON text, and they have no UTF-8 byte form.
  - Whether the catalog can express Track Names and Namespaces that are not valid UTF-8

_Assuming the catalog document itself is UTF8 on the wire, per [RFC 8259 8.1](https://www.rfc-editor.org/rfc/rfc8259#section-8.1)._ 

  ### Related

  - moq-wg/catalog-format (archived repo): moq-wg/catalog-format#45

wilaw:
- Updated track namespace and name definitions to use JSON arrays instead of strings. 
- Clarified constraints and requirements for track namespaces and names, including uniqueness and inheritance rules.
- Defines rules for how namespace and name fields are mapped to bytes.
- Updated all catalog examples.

Fixes: #209 
Fixes: [#45](https://github.com/moq-wg/catalog-format/issues/45)

### DIFF
--- a/draft-ietf-moq-msf.md
+++ b/draft-ietf-moq-msf.md
@@ -44,6 +44,7 @@ normative:
   MIME: RFC6838
   RFC9000: RFC9000
   GZIP: RFC1952
+  UTF8: RFC3629
   MOQLOG: I-D.draft-jennings-moq-log
   MOQMETRICS: I-D.draft-jennings-moq-metrics
   WEBCODECS-CODEC-REGISTRY:
@@ -444,16 +445,37 @@ is specified in Table 2.
 ### Track namespace {#tracknamespace}
 Required: Optional    JSON Type: String    Location: Track Object
 
-The name space under which the track name is defined. See section 2.3 of
-{{MoQTransport}}. The track namespace is optional. If it is not declared within
-a track, then each track MUST inherit the namespace of the catalog track. A
-namespace declared in a track object overrides any inherited name space.
+The Track Namespace under which the track name is defined. See Section 2.4.1 of
+{{MoQTransport}}. The value is a JSON array in which each element is a JSON
+string corresponding, in order, to one Track Namespace Field of the MOQT Track
+Namespace. Each element is converted to the bytes of its Track Namespace Field
+using the rules in {{namemapping}}.
+
+The array MUST contain between 1 and 32 elements, and every element MUST be a
+non-empty JSON string, so that the resulting Track Namespace satisfies the
+constraints of Section 2.4.1 of {{MoQTransport}}. A parser that encounters an
+empty array, more than 32 elements, an empty string element, or a non-string
+element MUST treat the catalog as invalid.
+
+The track namespace is optional. If it is not declared within a track object,
+then the track MUST inherit the Track Namespace of the catalog track, that is,
+the Track Namespace under which the catalog itself was subscribed. A namespace
+declared in a track object replaces (and does not extend) any inherited
+namespace.
 
 ### Track name {#trackname}
 Required: Yes    JSON Type: String    Location: Track Object
 
-A string defining the name of the track. See section 2.3 of {{MoQTransport}}.
-Within the catalog, track names MUST be unique per namespace.
+A JSON string defining the MOQT Track Name of the track. See Section 2.4.1 of
+{{MoQTransport}}. The string is converted to the bytes of the Track Name using
+the rules in {{namemapping}}. A Track Name MAY be the empty string.
+
+Within a catalog, the combination of Track Namespace and Track Name (the Full
+Track Name) MUST be unique. Uniqueness, and every other comparison of track
+names or namespaces required by this document (for example in {{dependencies}},
+{{parentname}}, {{parentnamespace}} and {{deltaupdate}}), is determined by exact
+comparison of the mapped bytes as described in {{namemapping}}, never by
+comparison of the JSON source text.
 
 ### Packaging {#packaging}
 Required: Yes    JSON Type: String    Location: Track Object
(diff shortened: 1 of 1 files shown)


draft-ietf-moq-msf.md
```diff
@@ -2464,6 +2497,100 @@ Note: This encoding ensures that the structural delimiters (- and --) remain una
...
+carry track names or namespaces MUST use the same mapping.
+
+### Producing bytes from a JSON string
+
+The MOQT byte value corresponding to a JSON string is the UTF-8 {{UTF8}}
+encoding of the sequence of Unicode scalar values obtained by parsing the JSON
+string per Section 7 of {{JSON}}. In particular:
+
+* JSON escape sequences are resolved before encoding. The string `"café"`
+  and the string `"café"` denote the same sequence of scalar values and
```
michalhosna, line 2517: > the string `"café"` and the string `"café"

Maybe you wanted to talk abou `\u` escaping here?

draft-ietf-moq-msf.md
```diff
@@ -2670,7 +2797,13 @@ entry with a type of 'object-property'.
...
 # Security Considerations
 
-ToDo
+Track names and namespaces are compared as raw bytes ({{namemapping}}). Because
+no Unicode normalization or case folding is applied, visually confusable names
+(for example, names differing only in composed versus decomposed accents, or in
+characters from different scripts that render identically) are distinct tracks.
+Catalog producers SHOULD avoid emitting distinct tracks whose names are
+confusable to a human reader, and applications that display track names to
+users SHOULD NOT rely on visual inspection to establish identity.
```
michalhosna, line 2806: Nitpick: I feel like there should be TODO left, as this is I suspect not the only security consideration?

draft-ietf-moq-msf.md
```diff
@@ -1789,7 +1817,7 @@ a different namespace.
     },
     {
       "name": "camera33-catalog",
-      "namespace": "example.com/blimps-R-us",
+      "namespace": ["example.com',"blimps-R-us"],
```
michalhosna, line 1820:
```suggestion
      "namespace": ["example.com","blimps-R-us"],
```

draft-ietf-moq-msf.md
```diff
@@ -1812,7 +1840,7 @@ track name formats follow the conventions defined in {{MOQLOG}} and {{MOQMETRICS
   "tracks": [
     {
       "name": "video",
-      "namespace": "broadcast.example.com/live/stream1",
+      "namespace": ["broadcast.example.com/","live","stream1"],
```
michalhosna, line 1843:
```suggestion
      "namespace": ["broadcast.example.com","live","stream1"],
```

draft-ietf-moq-msf.md
```diff
@@ -1826,7 +1854,7 @@ track name formats follow the conventions defined in {{MOQLOG}} and {{MOQMETRICS
     },
     {
       "name": "audio",
-      "namespace": "broadcast.example.com/live/stream1",
+      "namespace": ["broadcast.example.com/","live","stream1"],
```
michalhosna, line 1857:
```suggestion
      "namespace": ["broadcast.example.com","live","stream1"],
```

draft-ietf-moq-msf.md
```diff
@@ -1840,14 +1868,14 @@ track name formats follow the conventions defined in {{MOQLOG}} and {{MOQMETRICS
...
-      "namespace": "moq://metrics.moq.arpa/v1/%resourceId%",
+      "namespace": ["metrics.moq.arpa","v1","%resourceId%"],
       "name": "4",
       "packaging": "moqmetrics",
       "role": "metrics",
       "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9..."
     },
     {
-      "namespace": "moq://moq-syslog.arpa/logs-v1/%resourceId%",
+      "namespace": ["metrics.moq.arpa","v1","%resourceId%"],
```
michalhosna, line 1878: I am not sure what the change intention was, so just flagging that there is other-than-reformat change.
```suggestion
      "namespace": ["moq-syslog.arpa","v1","%resourceId%"],
```

draft-ietf-moq-msf.md
```diff
@@ -444,16 +445,37 @@ is specified in Table 2.
 ### Track namespace {#tracknamespace}
 Required: Optional    JSON Type: String    Location: Track Object
```
michalhosna, line 446:
```suggestion
Required: Optional    JSON Type: Array of String    Location: Track Object
```

draft-ietf-moq-msf.md
```diff
@@ -2166,16 +2195,13 @@ format specified in Section 4 of {{MOQLOG}}.
...
+The log track's namespace is defined by the {{tracknamespace}} field and converted to bytes
+according to {{namemapping}}.
 
-The Track Name is a single byte containing the log priority level in binary.
-Priority levels range from 0 (Emergency) to 7 (Debug), following syslog severity
-conventions.
+The Track Name is the ASCII decimal representation of the log priority level
+(for example, "6" for Informational), mapped to bytes as defined in
+{{namemapping}}. Priority levels range from 0 (Emergency) to 7 (Debug), following syslog
+severity conventions.
```
michalhosna, line 2204: Dropping compatibility with MOQLOG seems out of scope in this PR and little bit hidden. 

This should probably get it own Issue/PR. Sorry if it was already discussed/aggreed upon.

michalhosna (commented):
This does improve my issue, thanks. 

I am not sure if it is a good idea to use UTF-8 encoding this way, but I am not a text encoding expert and it seems sufficiently specified. 

Nitpicking: I feel the new text is very verbose / non-normative sounding.

gwendalsimon (approved):
I like it

